PR #54, opened 2018-03-15
virtual keyboard: Use push buttons instead of skinnables and fix focus object logic

peter-ha:
The virtual keyboard should behave like before, but it doesn't use the skinnables and neither its own skinlet anymore. Also, we fixed the focus object logic.


peter-ha:
Changes made since the first pull request a week ago:
* Layout updates (e.g. resizing etc.) are now working
* QskInputPanel has a Panel subcontrol to be styled
* Font size is growing on layout updates

PR #62, opened 2018-05-07
Speedometer: inherit from QskRangeControl and make skinnable

peter-ha:
... and display by default

### DIFF
--- a/examples/automotive/DefaultSkin.cpp
+++ b/examples/automotive/DefaultSkin.cpp
@@ -74,7 +74,7 @@ namespace {
         return font;
     }
 
-}
+} // namespace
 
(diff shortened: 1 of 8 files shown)


uwerat:
I inverted the order of the nodes to draw the needle below the box ( how is the thing called, where the needle is mounted ? ) and removed the code for the no longer used skins.

Would be good to have some sort of swipe/stack view, so that we can have both pages without having to use the SPEEDO compile switch.

peter-ha:
Ah yes it makes more sense to draw the needle head on top of the needle; I introduced an own NeedleHead subcontrol to style that individually.

I also think the next step would be to have some sort of swipe view to change between the sound control and speedometer pages...

PR #68, opened 2018-10-29
Do not include /usr/include with isystem

schnitzeltony:
gcc 8.2 fails with:

|  /usr/include/c++/8.2.0/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
|  #include_next <stdlib.h>
|                ^~~~~~~~~~

More details are found at [1]

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>

### DIFF
--- a/qskconfig.pri
+++ b/qskconfig.pri
@@ -62,7 +62,6 @@ linux {
         # to exclude them by declaring them as system includes
 
         QMAKE_CXXFLAGS += \
-            -isystem $$[QT_INSTALL_HEADERS] \
             -isystem $$[QT_INSTALL_HEADERS]/QtCore \
             -isystem $$[QT_INSTALL_HEADERS]/QtGui \
             -isystem $$[QT_INSTALL_HEADERS]/QtGui/$$[QT_VERSION]/QtGui \


uwerat:
Yocto makes QT_INSTALL_HEADERS being /usr/include - ah yes the default packages of my distro do the same.

The main problem with the Qt headers is, that they don't stand pedantic checks and flood all sort of code checking tools with warnings and errors. Simply removing QT_INSTALL_HEADERS from the include path will probably work with QSkinny, but only because it does not do any includes like this: #include <QtGui/QGuiApplication" - what is often common practice for writing Qt applications.

I will see how to solve the problem on the QSkinny side, but putting Qt subdirs to /usr/include has these obvious sort of disadvantages.

uwerat:
-I .. is what you get from the qmake specs, what unfortunately does not prevent the source code checkers from analyzing Qt includes. But as the specific path is not used by the QSkinny code it should be no problem to remove it.

I didn't apply your pull request for removing the offending include path, but your proposal is exactly what I did.

PR #74, opened 2019-03-06
Support building on macOS

richardmg:
Some changes I needed to do, to make it build and run on macOS. Basically every lib/plugin needs to set an install name, unless you want to twiddle with setting DYLD_LIBRARY_PATH etc. Also, since the skin plugins are linked in directly in the examples as a if they were normal libs, I needed to add _debug suffix to them when building in debug mode.

### DIFF
--- a/.qmake.conf
+++ b/.qmake.conf
@@ -1,0 +1,1 @@
+QMAKEFEATURES=$$PWD/features
--- a/features/default_post.prf
+++ b/features/default_post.prf
@@ -1,0 +1,7 @@
+load(default_post)
+
+darwin:equals(TEMPLATE, lib) {
+    # Add install name to all libraries, including plugins
+    LIBS += -Wl,-install_name,@rpath/lib$${TARGET}.dylib
+}
+


uwerat:
Thanks Richard,

this is very helpful as I don't have any Mac or Window systems and making things cross platform this way is always somehow like blind flying.

Maybe 2 questions:

a) darwin

Your patch shows rules for the darwin platform - what about other mac environments ?

b) qmake functions

In Qwt I am using a modified version of  qt_functions.prf: see https://sourceforge.net/p/qwt/code/HEAD/tree/trunk/qwt/qwtfunctions.pri. I would prefer to handle the rules you had to add to several project files in a similar way for qskinny.

Would you be able to come up with such a qsk_functions.pri or - if I do it - at least check some platforms I don' have ?

( Actually there are reports about build problems on the Mac for Qwt from time to time, where  I'm kind of clueless without having seen a Mac in my whole life )

uwerat:
I cleaned up the project files, so that everything is based on qtLibraryTarget and qtPlatformTargetSuffix and added several of your suggestions. What is missing is adding the -install_name linker flag.

Looking at the files in $QTDIR/mkspecs it seems, that all the pieces for this command can be found there:

QMAKE_LFLAGS_SONAME: -install_name
QMAKE_EXTENSION_SHLIB: dylib
QMAKE_SONAME_PREFIX: @rpath ( done in qt_module.prf )

QMAKE_LFLAGS_SONAME/QMAKE_EXTENSION_SHLIB are set in the related qmake.conf files, so the only one that needs to be added in the qskinny project files might be QMAKE_SONAME_PREFIX.

When also looking at the implementation in qmake/generators/unix/unixmake2.cpp it looks like qmake is fiddling some linker option together from these pieces.

So could you please try to only add "QMAKE_SONAME_PREFIX=@rpath" instead of the complete linker option - maybe this is already good enough to convince qmake doing the rest ?

richardmg:
After you new changes, everything now builds correctly :)

I can set QMAKE_SONAME_PREFIX, but the problem is that it doesn't have any effect on plugins. Which is why I ended up writing the linker option directly, to insert the name into the skin plugins as well. We could, as an alternative, use QMAKE_SONAME_PREFIX for normal libs, and then the linker line for plugins, but I'm not sure if we gain much. I have anyway updated the patch now, where I insert the install name in a default_post step instead, which reduces things quite a lot.

uwerat:
Trying your patch I noticed, that qskinny/features/default_post.prf is not processed in my build environment ( = using qmake + make from the shell ) ?

Actually I was not aware that I can "overload" certain prf files to add custom rules and when googling for "default_post.prf" I couldn't find any example making use of this mechanism. Do you have some information how this works ?

Concerning your patch: guess it would be more explicit to narrow the extra rule for plugins this way, right ?

plugin {
    darwin:equals(TEMPLATE, lib) {
        # Add install name to all libraries, including plugins
        LIBS += -Wl,-install_name,@rpath/lib$${TARGET}.dylib
    }
}

Would it make sense to create a change request for Qt to have this line for all plugins ? The QSkinny examples might not be the only projects, where the use of certain plugins is already known at link time.

richardmg:
Sorry for late reply, wanted to test building qskinny on my virtual ubuntu box before replying.
> Trying your patch I noticed, that qskinny/features/default_post.prf is not processed in my build environment
It runs on my virtual machine. Are you sure you did a rebuild? I added the line 'message("HERE")' in the default_post.prf, and it dumps it out when rebuilding the project.

But yeah, qmake docs are not the best :D Basically, it will look for default_pre.prf before it reads the project pro file, and then default_post.prf after processing the project pro file. The same goes for all the features you add to QT. E.g if you add 'Qt += foobar' in your pro file, qmake will look for a feature called foobar.prf. This can be used to e.g add a plugin.prf feature, so whenever you add 'QT += plugin', you get a chance to post process the build in the prf. QMAKE_PLATFORM values are also converted to features (e.g mac.prf, ios.prf). Some more info about the search path for prf files can be found [here](https://doc.qt.io/Qt-5/qmake-advanced-usage.html).

> Would it make sense to create a change request for Qt to have this line for all plugins ?
Perhaps. I'm not sure if you're supposed to be able to link in dynamic plugins as if they were a dynamic library in the first place. Do you know if this is normal practice elsewhere? Normally I guess you would build them as a static plugin for such cases?

uwerat:
> It runs on my virtual machine. Are you sure you did a rebuild? I added the line 'message("HERE")' in the default_post.prf, and it dumps it out when rebuilding the project.

Without doing some extra step - like  setting "export QMAKEFEATURES=.../qskinny/features' I can't get the prf file being processed. To my surprise I do not have to add "CONFIG += qskinny" to my project files, so there seem to be at least some sort of default processing rules being applied.

But I'm running qmake from the shell - maybe you have some sort of IDE ( Qt Creator ), that sets up a specific build environment ?

> I'm not sure if you're supposed to be able to link in dynamic plugins as if they were a dynamic library in the first place. Do you know if this is normal practice elsewhere? Normally I guess you would build them as a static plugin for such cases?

It simply makes sense to link as early as possible. When at application link time it is known, that you need a certain library you can improve the startup time + simplify the environment for your users.

Using static plugins might not be possible because of license implications or the simple fact, that plugin and application are from different projects and the project, where the plugin is coming from does not offer a static version.

In the specific situation of the qskinny examples one motivation was to avoid, that someone who wants to have a quick look at the project has to struggle with plugins not being loaded. But this has become obsolete after manually extending QT_PLUGIN_PATH in SkinnyPlugin.cpp.

The other reason is, that I want to show how skins can be implemented and assigned with and without using the plugin mechanism. In a real world scenario I know how I want to link my skins and wouldn't create it as plugin if I want to use it as a lib.

richardmg:
> Without doing some extra step - like setting "export QMAKEFEATURES=.../qskinny/features' I can't get the prf file being processed

Yeah, I had to set that too, but I do it in .qmake.conf, which is included in the patch. You add this file, right?

> But I'm running qmake from the shell - maybe you have some sort of IDE ( Qt Creator ), that sets up a specific build environment ?

Nope, I'm just using a command shell on ubuntu. No QtCreator, or any specific environment variables set.

I think it makes sense to be able to link in dynamic plugins like you do, so it would be nice qmake supported it out of the box. But I'm not sure if this can cause any regressions for others? I can discuss it with some other colleagues as well.

uwerat:
Indeed, I missed adding the .qmake.conf file  - I hope its syntax is not platform specific ?

To my surprise this works not only when building from the root directory, but also when building only being parts from a subdirectory - cool. So I also removed all the include directives from my local project files and added a default_pre.prf.

Thanks a lot - and please do a quick test of my changes.

PR #79, opened 2020-04-27
Fix cross compilation build for ARM

peter-ha:
* Add some INSTALLS directives to .pro files. They are used by
  QtCreator to deploy code to target devices, so it makes sense to
  deploy the examples and test support library as well.
* Add the skins plugins explicitly when adding test support;
  otherwise this won't build. This Might be a problem of the toolchain
  though.

### DIFF
--- a/examples/automotive/automotive.pro
+++ b/examples/automotive/automotive.pro
@@ -59,8 +59,13 @@ SVGSOURCES = \
 # writing application project files.
 ###########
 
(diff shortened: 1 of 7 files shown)


peter-ha:
Building svg2qvg for the host platform instead of the target platform is not addressed here

uwerat:
Parts of this patch have been applied. The problem with the missing svg2qvg tool, when cross compiling has been solved by adding the qvg files instead of building them.

PR #86, opened 2020-09-22
fixed build on Windows

vladimir-kraus:
There was an ambiguous symbol "Rectangle" on Windows, which caused that the project was not possible to build on Windows. Unfortunately the ambiguity was caused by including function Rectangle() from wingdi.h via QtQuick (i.e. outside of scope of QSkinny). So I guess the easiest solution was to rename the local symbol.

### DIFF
--- a/playground/grids/GridSkinny.cpp
+++ b/playground/grids/GridSkinny.cpp
@@ -10,10 +10,12 @@
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


uwerat:
Rectangle has been renamed to ColoredRectangle. Thanks for the patch.

PR #118, opened 2021-04-30
Don't replace saqt documentation tags

peter-ha:
It is easier for us to parse those special tags in Doxybook, too bad we cannot hook into the tags right in Doxygen.
The current version of the website already contains those commits.

### DIFF
--- a/doc/Doxyfile
+++ b/doc/Doxyfile
@@ -263,8 +263,7 @@ ALIASES                = "accessors=\par Access functions:^^" \
                          "subcontrols=\par Subcontrols:^^" \
                          "states=\par States:^^" \
                          "skinlet=\par Default Skinlet:^^" \
-                         "aspect=\par Aspect^^" \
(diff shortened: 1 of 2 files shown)


uwerat:
By simply removing the saqt alias the regular doxygen usage will be broken ?

peter-ha:
> By simply removing the saqt alias the regular doxygen usage will be broken ?

I see, do we need the doxygen-generated web page? In that case I need to check whether I can hack that somehow in doxybook2...

uwerat:
You could do an initial sed editing the Doxfile.

Another option might be to tweak doxyfilter.sh what is registered in Doxyfile. Currently it calls a tiny m4 script, but we could make it configurable by using environment variables ?

uwerat:
Of course doxyfilter.sh currently does not work as it is registered for *.h files only. But of course we can install another filter for *.dox files, that works in a similar way.

PR #122, opened 2021-08-02
QskSkinHintTableEditor: add overloads for all states

peter-ha:
... also, remove part of the logic when looking up current states.

### DIFF
--- a/qskinny.pro
+++ b/qskinny.pro
@@ -8,7 +8,8 @@ SUBDIRS = \
     tools \
     support \
     examples \
-    playground
+    playground \
+    tests
 
 OTHER_FILES = \
     doc/Doxyfile \
@@ -24,3 +25,4 @@ skins.depends = src
 support.depends = src skins
 examples.depends = tools support skins qmlexport
 playground.depends = tools support skins qmlexport
+tests.depends = src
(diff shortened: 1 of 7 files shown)


qskinny.pro
```diff
@@ -8,7 +8,8 @@ SUBDIRS = \
     tools \
     support \
     examples \
```
peter-ha, line 10: I could not find any example that make use of state combinations; the only example that uses states at all is the IOT dashboard. However that one sets different hints per state, so nothing we could make use of with the functions here.

src/controls/QskSkinHintTableEditor.h
```diff
@@ -30,9 +30,11 @@ class QSK_EXPORT QskSkinHintTableEditor
     // generic access
 
     void setHint( QskAspect, const QVariant& );
+    void setHint( QskAspect, const QVariant&, QskAspect::State );
```
peter-ha, line 33: Maybe the following would be more descriptive?

```suggestion
    void setHintForAllStates( QskAspect, const QVariant&, QskAspect::State );
```

tests/skinhints/skinhints.pro
```diff
@@ -0,0 +1,6 @@
+CONFIG += testcase qskinny
```
peter-ha, line 1: This line adds another make target ("make check"), which makes it handy to run all tests in a CI system (Github actions etc.)

peter-ha (commented):
I left some comments open for discussion

PR #123, opened 2021-08-05
Add thumb to SwitchButton

clemensmanert:


src/controls/QskSwitchButton.h
```diff
@@ -17,7 +17,7 @@ class QSK_EXPORT QskSwitchButton : public QskAbstractButton
         WRITE setInverted NOTIFY invertedChanged FINAL )
 
   public:
-    QSK_SUBCONTROLS( Groove, Handle )
+    QSK_SUBCONTROLS(Groove, Handle, Thumb)
```
uwerat, line 20: Groove/Handle is the terminology we used so far while Track/Thumb is what the material specs are using.
( I'm tempted to also use Track/Thumb in qskinny - using existing terminolgy is always better than to invent our own )

But having a Thumb + Handle subcontrols can't be correct. Obviously the new "Thumb" subcontrol is something different ?

clemensmanert: Ah I get it now. First I thought with thumb you where referring to the ripple shown in the states here https://material.io/components/switches#behavior . I created a subcontrol for that _Thumb_. Though it can be designed using the _Handle_'s border it is much easier this way. 

I can rename it into _Ripple_.

PR #136, opened 2021-10-20
Add arc borders and colors

peter-ha:


src/common/QskArcBorderColors.h
```diff
@@ -0,0 +1,90 @@
...
+#include <qmetatype.h>
+
+class QDebug;
+
+class QSK_EXPORT QskArcBorderColors
+{
+  public:
+    QskArcBorderColors();
+
+    QskArcBorderColors( const QColor& outer, const QColor& inner );
```
peter-ha, line 21: We could actually make this class a pair of gradients, then we could draw conical gradients with the QPainter API

src/nodes/QskArcRenderer.cpp
```diff
@@ -48,4 +52,28 @@ void QskArcRenderer::renderArc(const QRectF& rect,
...
+    if( borderMetrics.width( Qsk::Inner ) > 0 )
+    {
+        // draw inner border:
+        const qreal i = metrics.width() / 2;
+        const auto innerRect = rect.marginsRemoved( { i, i, i, i } );
+        const QColor innerColor = borderColors.color( Qsk::Inner );
+
+        painter->setPen( QPen( innerColor, borderMetrics.width( Qsk::Inner ),
+            Qt::SolidLine, Qt::FlatCap ) );
+        painter->drawArc( innerRect, metrics.startAngle(), metrics.spanAngle() );
```
peter-ha, line 65: We just draw additional arcs, because QPainter uses the pen to draw arcs, and there doesn't seem to be a way to draw borders with a pen. Otherwise we would have to call drawPie(), but that would introduce other difficulties.

src/common/QskNamespace.h
```diff
@@ -58,6 +58,13 @@ QSK_NAMESPACE( Qsk )
     };
     QSK_ENUM( Position )
 
+    enum ArcPosition
+    {
+        Outer,
+        Inner
+    };
+    QSK_ENUM( ArcPosition )
```
peter-ha, line 66: I didn't find an enum in Qt or QSkinny for this, so I introduced a new one

peter-ha:
Screenshot from 2021-10-19 14-12-33

PR #141, opened 2021-11-05
QskProgressBar: Don't hardcode orientation of gradient

peter-ha:
It makes sense to have a vertical gradient on a horizontal progress
bar etc.

### DIFF
--- a/src/controls/QskProgressBarSkinlet.cpp
+++ b/src/controls/QskProgressBarSkinlet.cpp
@@ -124,8 +124,6 @@ QSGNode* QskProgressBarSkinlet::updateBarNode(
     if ( !gradient.isVisible() )
         return nullptr;
 
-    gradient.setOrientation( bar->orientation() );
-
     if ( !gradient.isMonochrome() )
     {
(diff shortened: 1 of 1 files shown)


src/controls/QskProgressBarSkinlet.cpp
```diff
@@ -124,8 +124,6 @@ QSGNode* QskProgressBarSkinlet::updateBarNode(
     if ( !gradient.isVisible() )
         return nullptr;
 
-    gradient.setOrientation( bar->orientation() );
-
```
uwerat, line 128: This line was made for a situation, where the gradient is in direction of the progress ( f.e cold/hot -> blue/red ) without having to specify it twice using the placement bits. One of the situations, where we would need something like a QskGradient::NoOrientation or having QskColorStops only.

I recently played with extending the QskGradient::Orientation enum, but was not yet happy with all consequences in other places. Need to think more about it ...

peter-ha: I can see the need for this and a new enum. Will put this PR on hold, as it is not that easy...

PR #144, opened 2021-11-11
Improve material theme

clemensmanert:
This pr tries to match more material design rules. 
- light and dark themes
- one and two color themes
- material style focus
- support for disabled elements
- sane defaults

There are still some things that are  missing:
- shadows, this is easy to add once they are stable in skinny
- support for different layers, where upper layers are lighter/darker than those below
- semantic elements, like different styles for buttons according to their purpose 
- overlaying scrollbars in listviews 
- control specific elements, like labels inside text inputs
- ripple animations on a surface (https://material.io/design/interaction/states.html#usage)

All but the first of these points require subclassing the controls in order to get closer toe the materials specs. 

This pr also removes the ripple from  QskSwitchButton, since the same effect can be also archived using borders. 

Edit: Added ripple animations

### DIFF
--- a/skins/material/QskMaterialSkin.cpp
+++ b/skins/material/QskMaterialSkin.cpp
@@ -37,7 +37,6 @@
 #include <QskFunctions.h>
 #include <QskMargins.h>
 #include <QskNamespace.h>
-#include <QskRgbValue.h>
 
 #if 1
 // should be defined in the public header, so that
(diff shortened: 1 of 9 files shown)


uwerat:
Could you please come up with 2 patches: one for the changes of QskSwitchButton and the other for the theming ?

clemensmanert:
Sure

clemensmanert:
Sure

PR #146, opened 2021-11-16
QskTabBar: Respect alignment hint

peter-ha:
The only solution I could find was to add spacers next to the button box. For some reason setting the extra space did not work, because the buttons would always take up the extra space, even when setting the size policy to fixed.


uwerat:
QskTabBar will be reimplemented from scratch taking a similar approach to QskSegmentedBar. Maybe we do not need QskTabBar at all - having an all-in-one QskTabView might be good enough.

So this pull request has become obsolete

PR #158, opened 2022-01-10
Box renderer: Support border gradients

peter-ha:


peter-ha:
Screenshot of the adapted boxes example:

Screenshot from 2022-01-17 07-05-15

src/common/QskBoxBorderColors.cpp
```diff
@@ -15,44 +15,47 @@ static void qskRegisterBoxBorderColors()
...
 
 Q_CONSTRUCTOR_FUNCTION( qskRegisterBoxBorderColors )
 
-static inline bool qskIsVisble( const QColor& c )
-{
-    return c.isValid() && ( c.alpha() > 0 );
-}
-
-static inline void qskSetColors( const QColor& c, QColor* colors )
+static inline void qskSetGradients( const QskGradient& c, QskGradient* gradients )
```
uwerat, line 25: c -> gradient

src/common/QskBoxBorderColors.cpp
```diff
@@ -61,105 +64,104 @@ QskBoxBorderColors::~QskBoxBorderColors()
...
-        m_colors[ Qsk::Right ] = c;
+        m_gradients[ Qsk::Right ] = gradient;
 
     if ( edges & Qt::BottomEdge )
-        m_colors[ Qsk::Bottom ] = c;
+        m_gradients[ Qsk::Bottom ] = gradient;
 }
 
-QColor QskBoxBorderColors::colorAt( Qt::Edge edge ) const
+QskGradient QskBoxBorderColors::gradientAt( Qt::Edge edge ) const
```
uwerat, line 113: const QskGradient& - a copy is not too expensive, as the stops are a QVector, but as this class is used in the box renderer I would avoid it. For the situation, where no edge has been selected I would write:

    static QskGradient noGradient;
    return noGradient;

src/nodes/QskBoxNode.cpp
```diff
@@ -117,7 +117,7 @@ void QskBoxNode::setBoxData( const QRectF& rect,
     {
         if ( isFillMonochrome && isBorderMonochrome )
         {
-            if ( borderColors.color( Qsk::Left ) == fillGradient.startColor() )
+            if ( borderColors.gradient( Qsk::Left ) == fillGradient.startColor() )
```
uwerat, line 120: We have to compare the startColors only - not the gradients itself.
Otherwise we might fail with the orientation of the gradient !

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -532,8 +546,29 @@ namespace
...
+                    yEnd = y12 + p * ( y22 - y12 );
+
+                lines[ additionalStopCount - i + 1 ].setLine( xStart, yStart,
+                    xEnd, yEnd, s.at( i ).color() );
+            }
+        }
+
         template< class BorderMap, class FillMap >
-        inline void createLines( Qt::Orientation orientation, Line* borderLines,
+        void createLines( Qt::Orientation orientation, Line* borderLines,
```
uwerat, line 571: Why removing the inline keyword ?

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -641,6 +682,91 @@ namespace
...
                     }
+
+                    // at the beginning and end of the loop we can add
+                    // additional lines for border gradients:
+
+                    if( j == 0 )
+                    {
+                        if( additionalGradientStops( borderMapTR.gradient() ) > 0 )
+                        {
+                            auto stops = borderMapTR.gradient().stops();
```
uwerat, line 693: const auto stops

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -641,6 +682,91 @@ namespace
...
+                            y2TL = c[ TopLeft ].centerY - v.dy2( TopLeft );
+
+                            addAdditionalLines( x1TR, y1TR, x2TR, y2TR,
+                                                x1TL, y1TL, x2TL, y2TL,
+                                                borderMapTR.gradient(), linesTR + k );
+                        }
+
+                        if( additionalGradientStops( borderMapBL.gradient() ) > 0 )
+                        {
+                            auto stops = borderMapBL.gradient().stops();
```
uwerat, line 712: const auto stops

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -641,6 +682,91 @@ namespace
...
+                                                x1BR, y1BR, x2BR, y2BR,
+                                                borderMapBL.gradient(), linesBL + k );
+                        }
+                    }
+
+                    if( j == numCornerLines - 1 )
+                    {
+                        if( additionalGradientStops( borderMapTL.gradient() ) > 0 )
+                        {
+                            auto stops = borderMapTL.gradient().stops();
```
uwerat, line 734: const auto stops

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -641,6 +682,91 @@ namespace
...
+                            y2BL = c[ BottomLeft ].centerY + v.dy2( BottomLeft );
+
+                            addAdditionalLines( x1TL, y1TL, x2TL, y2TL,
+                                                x1BL, y1BL, x2BL, y2BL,
+                                                borderMapTL.gradient(), linesTL + j );
+                        }
+
+                        if( additionalGradientStops( borderMapBR.gradient() ) > 0 )
+                        {
+                            auto stops = borderMapBR.gradient().stops();
```
uwerat, line 753: const auto& stops

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -859,17 +990,24 @@ static inline void qskRenderBorder( const QskBoxRenderer::Metrics& metrics,
 
     if ( colors.isMonochrome() )
     {
-        qskRenderBorderLines( metrics, orientation, line, BorderMapSolid( c.rgb( Qsk::Left ) ) );
+        qskRenderBorderLines( metrics, orientation, line, BorderMapSolid( c.gradient( Qsk::Left ).startColor().rgb() ) );
```
uwerat, line 993: rgb -> rgba

To avoid running into this trap I would introduce something like this at the top of this file:

static inline QRgb qskRgbGradient( const QskGradient& gradient )
{
    return gradient.startColor().rgba();
}

static inline QRgb qskRgbBorder( const QskBoxBorderColors& borderColors )
{
    return qskRgbGradient( borderColors.gradient( Qsk::Left ) );
}

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -859,17 +990,24 @@ static inline void qskRenderBorder( const QskBoxRenderer::Metrics& metrics,
...
 
+        auto left = c.gradient( Qsk::Left ), top = c.gradient( Qsk::Top ),
+                right = c.gradient( Qsk::Right ), bottom = c.gradient( Qsk::Bottom );
+
         qskRenderBorderLines( metrics, orientation, line,
-            BorderMapGradient( stepCount, c.rgb( Qsk::Top ), c.rgb( Qsk::Left ) ),
-            BorderMapGradient( stepCount, c.rgb( Qsk::Right ), c.rgb( Qsk::Top ) ),
-            BorderMapGradient( stepCount, c.rgb( Qsk::Left ), c.rgb( Qsk::Bottom ) ),
-            BorderMapGradient( stepCount, c.rgb( Qsk::Bottom ), c.rgb( Qsk::Right ) ) );
+            BorderMapGradient( stepCount, top.startColor().rgb(),
```
uwerat, line 1003: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -859,17 +990,24 @@ static inline void qskRenderBorder( const QskBoxRenderer::Metrics& metrics,
...
+                               left.endColor().rgb(), left ),
+            BorderMapGradient( stepCount, right.startColor().rgb(),
```
uwerat, line 1005: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -859,17 +990,24 @@ static inline void qskRenderBorder( const QskBoxRenderer::Metrics& metrics,
...
+                               top.endColor().rgb(), top ),
+            BorderMapGradient( stepCount, left.startColor().rgb(),
```
uwerat, line 1007: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -859,17 +990,24 @@ static inline void qskRenderBorder( const QskBoxRenderer::Metrics& metrics,
...
+                               bottom.endColor().rgb(), bottom ),
+            BorderMapGradient( stepCount, bottom.startColor().rgb(),
```
uwerat, line 1009: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -899,7 +1037,7 @@ static inline void qskRenderBoxRandom(
 
     if ( bc.isMonochrome() )
     {
-        const BorderMapSolid borderMap( bc.rgb( Qsk::Left ) );
+        const BorderMapSolid borderMap( bc.gradient( Qsk::Left ).startColor().rgb() );
```
uwerat, line 1040: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -918,10 +1056,18 @@ static inline void qskRenderBoxRandom(
     {
         const int n = metrics.corner[ 0 ].stepCount;
 
-        const BorderMapGradient tl( n, bc.rgb( Qsk::Top ), bc.rgb( Qsk::Left ) );
-        const BorderMapGradient tr( n, bc.rgb( Qsk::Right ), bc.rgb( Qsk::Top ) );
-        const BorderMapGradient bl( n, bc.rgb( Qsk::Left ), bc.rgb( Qsk::Bottom ) );
-        const BorderMapGradient br( n, bc.rgb( Qsk::Bottom ), bc.rgb( Qsk::Right ) );
+        const BorderMapGradient tl( n, bc.gradient( Qsk::Top ).startColor().rgb(),
```
uwerat, line 1059: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -918,10 +1056,18 @@ static inline void qskRenderBoxRandom(
...
+                                    bc.gradient( Qsk::Left ).endColor().rgb(),
+                                    borderColors.gradient( Qsk::Left ) );
+        const BorderMapGradient tr( n, bc.gradient( Qsk::Right ).startColor().rgb(),
```
uwerat, line 1062: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -918,10 +1056,18 @@ static inline void qskRenderBoxRandom(
...
+                                    bc.gradient( Qsk::Top ).endColor().rgb(),
+                                    borderColors.gradient( Qsk::Top ) );
+        const BorderMapGradient bl( n, bc.gradient( Qsk::Left ).startColor().rgb(),
```
uwerat, line 1065: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -918,10 +1056,18 @@ static inline void qskRenderBoxRandom(
...
+                                    bc.gradient( Qsk::Bottom ).endColor().rgb(),
+                                    borderColors.gradient( Qsk::Bottom ) );
+        const BorderMapGradient br( n, bc.gradient( Qsk::Bottom ).startColor().rgb(),
```
uwerat, line 1068: rgb -> rgba

src/nodes/QskBoxRendererEllipse.cpp
```diff
@@ -1124,7 +1270,7 @@ void QskBoxRenderer::renderRectellipseBorder(
     const int stepCount = metrics.corner[ 0 ].stepCount;
     const int lineCount = 4 * ( stepCount + 1 ) + 1;
 
-    const auto line = allocateLines< Line >( geometry, lineCount );
+    const auto line = allocateLines< ColoredLine >( geometry, lineCount );
```
uwerat, line 1273: Changing Line to ColoredLine looks wrong to me.

When having a QSGFlatColorMaterial there is only one color that is set with QSGFlatColorMaterial::setColor ( see QsBoxNode::setBoxData) and the corresponding geometry is using vertexes without extra color information ( running into a different shader then ): see QskBoxNode::setMonochrome.

src/nodes/QskBoxRendererRect.cpp
```diff
@@ -422,28 +422,58 @@ static inline void qskCreateBorder(
...
+    const QskGradient gradientTop = colors.gradient( Qsk::Top );
+    const QskGradient gradientBottom = colors.gradient( Qsk::Bottom );
 
-    ( line++ )->setLine( in.right, in.bottom, out.right, out.bottom, colorBottom );
-    ( line++ )->setLine( in.left, in.bottom, out.left, out.bottom, colorBottom );
+    // qdebug
 
-    if ( colorLeft != colorBottom )
-        ( line++ )->setLine( in.left, in.bottom, out.left, out.bottom, colorLeft );
+    for( const QskGradientStop& stop : gradientBottom.stops() )
```
uwerat, line 432: const auto&

src/nodes/QskBoxRendererRect.cpp
```diff
@@ -422,28 +422,58 @@ static inline void qskCreateBorder(
...
+    {
+        const Color c( stop.color() );
+        const qreal x1 = in.right - stop.position() * ( in.right - in.left );
```
uwerat, line 435: ( in.right - in.left ) and ( out.right - out.left ) could be precalulated outside of the loop. A good compiler should do this on its own, but to be sure:

{
    const qreal y1 = in.bottom;
    const qreal y2 = out.bottom;
    const qreal dx1 = in.right - in.left;
    const qreal dx2 = out.right - out.left

    for( const QskGradientStop& stop : gradientBottom.stops() )
    {
    }
}

src/nodes/QskBoxRendererRect.cpp
```diff
@@ -461,7 +491,7 @@ void QskBoxRenderer::renderRectBorder(
         return;
     }
 
-    const auto line = allocateLines< Line >( geometry, 4 + 1 );
+    const auto line = allocateLines< ColoredLine >( geometry, 4 + 1 );
```
uwerat, line 494: Looks wrong to me: see my comments about QSGFlatColorMaterial above

src/nodes/QskBoxRendererRect.cpp
```diff
@@ -587,7 +618,7 @@ void QskBoxRenderer::renderRect(
         auto fillLines = line + fillLineCount;
 
         if ( bc.isMonochrome() )
-            qskCreateBorder( rect, in, bc.rgb( Qsk::Left ), fillLines );
+            qskCreateBorder( rect, in, bc.gradient( Qsk::Left ).startColor().rgb(), fillLines );
```
uwerat, line 621: rgb -> rgba

uwerat (changes requested):

src/common/QskBoxBorderColors.cpp
peter-ha: You meant that the method should return a const ref, right? If yes, then the code is fine now.

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: Qt has this rule of thumb where you should only inline methods that have ~10 lines or so, otherwise the code gets too big. Since we don't use this function that often, I have reintroduced the inline keyword again.

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: It turns out we didn't use the variable at all, so I removed them.

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: dito

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: dito

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: dito

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: I introduced those methods above and one qskRgbGradientEnd() for the RGBA of the end color, since we need this function as well.

src/nodes/QskBoxRendererEllipse.cpp
peter-ha: Yeah, this turned out to be just an oversight I made when I started working on this.

uwerat:
Good work !

PR #159, opened 2022-01-19
Hunspell: Move prediction to an own thread

peter-ha:


src/inputpanel/QskInputContext.cpp
```diff
@@ -494,18 +500,62 @@ void QskInputContext::commitPrediction( bool )
      */
 }
 
+class QskInputContextFactory::PrivateData
+{
+  public:
+    QThread* thread = nullptr;
```
peter-ha, line 506: Maybe we should move the thread handling to the input context?

src/inputpanel/QskInputContext.cpp
```diff
@@ -494,18 +500,62 @@ void QskInputContext::commitPrediction( bool )
...
+    std::shared_ptr< QskTextPredictor > predictor;
```
peter-ha, line 507: I made this a shared pointer to support the case where an input panel outlives the predictor (e.g. when an input panel is still open and somebody switches the language and thus deletes the current predictor).
Also, I guess it is debatable whether we should have the methods related to the predictor in the input context factor; the naming sounds a bit misleading...

PR #163, opened 2022-02-18
Improve material theme

clemensmanert:


uwerat:
Started with understanding and piecewise merging the branch. Hope I will have everything in master in a couple of days.

uwerat:
we will go with Material 3

PR #81, opened 2020-06-04
initial github build action

oracle2025:
Building QSkinny with Github Actions Continuous Integration

PR #175, opened 2022-04-12
github build action for Linux and Windows

peter-ha:
This extends https://github.com/uwerat/qskinny/pull/81 by using Qt 5.15 and building for both Linux and Windows.

@uwerat : When merging, could you please rebase and leave the commits as they are, so that @oracle2025's commit is still recognizable as his own?

### DIFF
--- a/.github/workflows/qmake.yml
+++ b/.github/workflows/qmake.yml
@@ -1,0 +1,21 @@
+name: CI
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


peter-ha:
This is already live at https://github.com/qskinny/qskinny/actions, right now there is a real build error on Windows, while on Linux the build works.

PR #188, opened 2022-05-02
Windows: Fix placement policy of all items

peter-ha:
Before, on Windows nothing would be displayed because of the
following error:

implicit truncation from 'QskPlacementPolicy::Policy' to bit-field
changes value from 2 to -2

### DIFF
--- a/src/common/QskPlacementPolicy.h
+++ b/src/common/QskPlacementPolicy.h
@@ -50,8 +50,8 @@ class QskPlacementPolicy
     constexpr Policy hiddenPolicy() const noexcept;
 
   private:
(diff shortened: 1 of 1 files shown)


peter-ha:
before:
Screenshot (3)

after:
Screenshot (1)

uwerat:
Interesting, that we do not have the same issue on linux.

uwerat:
Looks like we have smoke test errors with Qt6 on Windows now ?

peter-ha:
> Looks like we have smoke test errors with Qt6 on Windows now ?

strange, I'll check it out tomorrow...

PR #190, opened 2022-05-04
CI: Use OpenGL backend on Windows

peter-ha:
We currently don't support the D3D11 backend, since we rely on
OpenGL functionality when rendering textures, see e.g.
qskCreateTextureRaster().

### DIFF
--- a/.github/workflows/qmake.yml
+++ b/.github/workflows/qmake.yml
@@ -244,6 +244,8 @@ jobs:
   build-windows-qt-lts:
     name: Windows Qt 6.2 (LTS) build
     runs-on: windows-latest
+    env:
+      QSG_RHI_BACKEND: opengl
     steps:
     - uses: actions/checkout@v2
     - name: Cache Qt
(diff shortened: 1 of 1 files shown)


peter-ha:
This makes the CI pass, but not the app start correctly. The problem is that Github Actions is using Windows Server, which doesn't come with the needed OpenGL libraries. On Windows Professional this works.

Now the next step is for the Windows CI will be to install Mesa OpenGL drivers...

uwerat:
Or to get rid of native OpenGL. Using QSGTexture instead of texture ids would probably remove most of the OpenGL related issues

peter-ha:
> Or to get rid of native OpenGL. Using QSGTexture instead of texture ids would probably remove most of the OpenGL related issues

Sure, that would be the even better solution, but would take some more effort I guess. Could we get rid of all the other direct OpenGL calls in QskTextureRenderer::qskCreateTextureRaster() though?

uwerat:
Sure - there are Qt APIs available to upload a QImage as texture, that I could have used in the first place. The only reason I implemented my own code was, that I wanted to have it more lightweight. Back in the days creating QObjects ( QSGTexture is one ) was something I tried to avoid for "marketing" reasons as the number of QObjects/QQuickItems is an indicator for the heaviness of an implementation. And because of the scalability and color manipulations QSkinny creates/destroys textures way more often than standard Qt/Quick applications.

But with the QRHI abstraction this is probably not worth the effort. It might be even possible to get the 2D renderer working for textures then - let's see ...

PR #193, opened 2022-05-27
QskPushButton: respect the graphics alignment

peter-ha:
Also, respect the padding hint of the graphic

### DIFF
--- a/src/controls/QskPushButtonSkinlet.cpp
+++ b/src/controls/QskPushButtonSkinlet.cpp
@@ -77,11 +77,24 @@ QRectF QskPushButtonSkinlet::textRect(
 
     if ( button->hasGraphic() )
     {
-        // in case of having text + graphic we put the text at the bottom
+        const auto orientation = button->flagHint( QskPushButton::Panel | QskAspect::Direction, Qsk::LeftToRight );
 
-        qreal h = button->effectiveFontHeight( QskPushButton::Text );
-        if ( h < r.height() )
-            r.setTop( r.bottom() - h );
+        switch( orientation )
+        {
+            case Qsk::LeftToRight:
+            {
+                const auto graphicsRect = subControlRect( button, contentsRect, QskPushButton::Graphic );
+                const auto spacing = button->metric( QskPushButton::Panel | QskAspect::Spacing );
+                r.setX( r.x() + graphicsRect.width() + spacing );
+                break;
+            }
+            default: // Qsk::TopToBottom, the other ones are not handled yet
+            {
+                qreal h = button->effectiveFontHeight( QskPushButton::Text );
+                if ( h < r.height() )
+                    r.setTop( r.bottom() - h );
+            }
+        }
     }
 
     return r;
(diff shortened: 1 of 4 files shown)


peter-ha:
Screenshot from 2022-05-27 14-18-06

See above. I guess if we wanted to use the Placement instead of the Alignment we would have to add some functions to QskSkinHintTableEditor, since this setting is per skin?

uwerat (changes requested):
I wouldn't use the alignment flags of the graphic for the desired feature as it is (a) not too intuitive and (b) blocks the flag to be used for how to align the icon inside of its given space.

Maybe it's better to use Qt::Orientation or Qt::LayoutDirection. Not 100% sure if we would have to respect QskQuickItem::layoutMirroring() ( might depend on the locale ) for the horizontal layout situation. Usually the buttons to control dialogs have a horizontal orientation ( XFCE4, Gnome, --- ). Might be interesting to check what happens to icon/text, when setting a locale, where the layout direction is from right to left ( f.e Hebrew )

Then the alignment can be used to define how graphic/text are aligned inside of their given rectangle. F.e when having a vertical layout and the text is longer than the width of the graphic you might want to have the graphic aligned to the left. Or f.e when having no graphic you might want to have the text at the bottom.

src/controls/QskPushButtonSkinlet.cpp
```diff
@@ -77,11 +77,24 @@ QRectF QskPushButtonSkinlet::textRect(
...
+            case Qt::AlignLeft:
+            {
+                const auto graphicsRect = subControlRect( button, contentsRect, QskPushButton::Graphic );
+                const auto spacing = button->metric( QskPushButton::Panel | QskAspect::Spacing );
+                r.setX( r.x() + graphicsRect.width() + spacing );
+                break;
+            }
+            default: // AlignTop
+            {
+                qreal h = QFontMetricsF( button->effectiveFont( QskPushButton::Text ) ).height();
```
uwerat, line 93: using effectiveFontHeight() ?

uwerat (commented):
Reviewed

peter-ha:
> I wouldn't use the alignment flags of the graphic for the desired feature as it is (a) not too intuitive and (b) blocks the flag to be used for how to align the icon inside of its given space.

I understand the point that we want to use the alignment flag for aligning the graphic inside its rectangle.

> 
> Maybe it's better to use Qt::Orientation or Qt::LayoutDirection. Not 100% sure if we would have to respect QskQuickItem::layoutMirroring() ( might depend on the locale ) for the horizontal layout situation. Usually the buttons to control dialogs have a horizontal orientation ( XFCE4, Gnome, --- ). Might be interesting to check what happens to icon/text, when setting a locale, where the layout direction is from right to left ( f.e Hebrew )

So are you suggesting just adding a getter and setter for an orientation, just like e.g. the segmented bar has? I can see how that would work, but what I am still wondering is:
We would not be able to style that in a skin; e.g. for Material all button graphics should be left of the text. Should we just accept that we would have to set this manually per button instance, and e.g. have an overload for the constructor?
I am not sure it would be intuitive to have a horizontal and vertical push button though...

peter-ha:
implemented with orientation:

Screenshot from 2022-06-14 16-06-46

uwerat (changes requested):
I was actually thinking about setting the orientation as a skin hint - then you can define the define the default value in the skin, what is needed for Material, as you already pointed out.

Offering the orientation as a property in the public API of QskPushButton is another discussion. But changing the orientation in application code sounds like an exotic use case to me and as it can always be done with:

    `button->setFlagHint( QskPushButton::Panel | QskAspect::Direction, Qsk::LeftToRight )`

I would have a tendency towards not having it as property.

IIRC we had more primitives before ( including QskAspect::Direction ? ) that have been removed in a cleanup session. In general we have 5 bits ( today ) for the primitives, what allows to define 15 more ( even more possible as it only has to be unique in combination with the type ).

PR #197, opened 2022-06-14
Material theme: Move to Material 3

peter-ha:


peter-ha:
> Is it possible to split this merge request - the first part with the changes according to new type of palette only ?

I stripped this out at https://github.com/uwerat/qskinny/pull/199 . I guess if this PR is merged, merging it back into this branch here will create quite a mess. I think the best way to deal with this branch here is to either merge it soon or try to copy the relevant stuff out into a new branch, without the history.

uwerat:
Please rebase and adjust the code using "color tones" calculated with QskHctColor.h

PR #199, opened 2022-06-15
Material: Switch palettes to Material 3

peter-ha:
Also, remove the old Material 2 palettes, since the weights etc.
changed.


uwerat:
With the implementation of the HCT algo the changes have become obsolete. Please rebase.

PR #202, opened 2022-06-15
Popups: Remove clip when showing the popup

peter-ha:
Otherwise we would clip the shadow, e.g. in the Material case.


uwerat:
Actually clipping is only necessary at the top border and therefore I changed the clip rectangle.
I also disabled the clip when the menu is fully visible - clipping breaks batching.

However, when there is a shadow at the top edge it would be clipped off while sliding in.

PR #203, opened 2022-06-16
Segmented bar: Allow for styling them for Material 3

peter-ha:
Screenshot from 2022-06-17 12-53-02


peter-ha:
@uwerat : In case you consider those commits fine to go to master, I think it makes sense to leave them separate by choosing "Rebase and merge" instead of "Squash and merge".

PR #212, opened 2022-08-02
Add cube effect to QskStackboxAnimator

peter-ha:
The effect works in all directions in the layouts example; however when implementing it in the IOT example, I get strange effects, see the screenshot:

Screenshot from 2022-08-02 11-58-09

@uwerat : Do you have any idea what might cause this?


uwerat:
Looks like I'm blind - what is wrong in the screenshot ?

peter-ha:
> Looks like I'm blind - what is wrong in the screenshot ?

The arcs and svgs are not displayed anymore; sometimes they are also just slightly off, in which case it is more visible.

peter-ha:
I just checked: With the QskStackboxAnimator2 (flipable) the effects are the same; I just commented out the effect in the IOT example for now.

PR #219, opened 2022-08-31
Material 3: Calculate real pixels from device-independent pixels

### LINKED ISSUE #210: Material3 and device independent pixels
In the Material specs we find metrics given in "dp" ( f.e icons as 24dp -> https://material.io/design/iconography/system-icons.html#grid-and-keyline-shapes ). "dp" stands for density independent pixels: https://en.wikipedia.org/wiki/Device-independent_pixel

According to https://developer.android.com/training/multiscreen/screendensities#TaskUseDP the formula for translating dp to device depending pixels seems to be: px = dp * (dpi / 160)

I created corresponding C++ user literals in QskMaterial3Skin.cpp and used them for the strutSize of the push button icons.

The calculatation in the literal needs to be verified and probably all values for the metrics need to be given in dp ?

peter-ha:
Resolves #210

### DIFF
--- a/skins/material3/QskMaterial3Skin.cpp
+++ b/skins/material3/QskMaterial3Skin.cpp
@@ -45,8 +45,8 @@
 #include <QskNamespace.h>
 #include <QskPlatform.h>
 
-#include <qguiapplication.h>
-#include <qscreen.h>
+#include <QGuiApplication>
+#include <QScreen>
 
 static const int qskDuration = 150;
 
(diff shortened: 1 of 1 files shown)


peter-ha:
LDPI:

material-ldpi

MDPI:

material-mdpi


HDPI:

material-hdpi


LDPI still looks a bit small, so we could also get rid of it. At least it is visible that with HDPI the buttons and fonts are larger than in the browser window.

Issue #220, opened 2022-09-01
Add blurring feature


rick-vogel:
# What

As a developer I want to make a **rounded rectangle's** background **blurred**. Furthermore I want to define the blurring algorithms parameters.

# How

- Add the blurring effect as a new shader

uwerat:
See https://github.com/uwerat/qskinny/tree/features/effectnode/playground/parrots

PR #221, opened 2022-09-01
proof of concept for #220

rick-vogel:
@uwerat hier mal der PoC für das Blurring. Achtung: Qt6 noch nicht integriert!


examples/blurredbox/MainWindow.cpp
```diff
@@ -0,0 +1,205 @@
...
+            case BL:
+                shape.setBottomLeft( { value, value } );
+                break;
+        }
+
+        label->setText( format.arg( value ) );
+        blurred->setShape( shape );
+        blurred->update();
+    } );
+    connect( bar, &QskSegmentedBar::currentIndexChanged, [ blurred, slider ]( int index ) {
```
uwerat, line 173: See https://github.com/KDE/clazy/blob/master/docs/checks/README-connect-3arg-lambda.md.

In this particular case the following call would be less error prone:
connect( bar, &QskSegmentedBar::currentIndexChanged, slider. ... );

However you would still have the lifecycle of blurred being unguarded. Lambdas as slots do have their pitfalls.

examples/blurredbox/main.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+int main( int argc, char* argv[] )
+{
+#ifdef ITEM_STATISTICS
+    QskObjectCounter counter( true );
+#endif
+
+    QGuiApplication app( argc, argv );
+
+    qskSetup->setItemUpdateFlag( QskQuickItem::PreferRasterForTextures, true );
```
uwerat, line 22: Pobably not needed.

uwerat:
Guess the main unsolved problem is to know when to update the scene graph of the blurred background.

When the item is not a child of the root item of the window its scene position might change without getting a notification from the system ( the coordinates relative to the parent might be unchanged ). The other situation is when the background changes on its own - f.e when the blurred box is on top of a camera.

rick-vogel:
Yes, that's a good point. Didn't consider this scenario. Unfortunately I also discovered another issue: if a text item (QskTextLabel or QskPushButton) is "behind" the blurred box the shader texture is "garbage"

image

uwerat:
If you have problems with text nodes you will probably have problems with all sort of nodes with vertex based geometries.

examples/blurredbox/shaders/blurredbox.frag
```diff
@@ -0,0 +1,84 @@
...
+    // Pixel colour
+    vec4 fragColor = texture2D(txr, coord);
+
+    // Blur calculations
+    const lowp float Pi = 6.28318530718; // constant for Pi*2
+    for( float d = 0.0; d < Pi; d += Pi / blurDirections)
+    {
+        for(float i = 1.0 / blurQuality; i <= 1.0; i += 1.0 / blurQuality)
+        {
+            fragColor += texture2D( txr, coord + vec2(cos(d), sin(d)) * blurRadius * i);
```
uwerat, line 71: cos/sin can be calculated outside of the inner loop. Actually it would good enough to calculate only cos and use Pythagoras for the sin ( or v.v. ) then. Can imagine that this will speed up the shader significantly. Working with precalculated tables for certain values of blurDirections should be make things even faster.

This shader also seems to do all the expensive blur calculations for pixels that will be sorted out later.

But of course things have to work first before doing performance improvements,

rick-vogel:
> Guess the main unsolved problem is to know when to update the scene graph of the blurred background.
> 
> When the item is not a child of the root item of the window its scene position might change without getting a notification from the system ( the coordinates relative to the parent might be unchanged ). The other situation is when the background changes on its own - f.e when the blurred box is on top of a camera.

I did some research in the Qt source code and found some interesting classes:

- to know when to update the scene graph => the [&QQuickItem::childrenChanged](https://github.com/qt/qtgraphicaleffects/blob/5.15.2/src/effects/private/qgfxsourceproxy.cpp#L76) signal in QQuickItem used here sounds promising, unfortunately it is undocumented

vgl.

```cpp
    if (m_input != nullptr) {
        if (QQuickImage *image = qobject_cast<QQuickImage *>(m_input)) {
            connect(image, &QQuickImage::sourceSizeChanged, this, &QGfxSourceProxy::repolish);
            connect(image, &QQuickImage::fillModeChanged, this, &QGfxSourceProxy::repolish);
        }
        connect(m_input, &QQuickItem::childrenChanged, this, &QGfxSourceProxy::repolish);
    }
```

uwerat:
QQuickItem::childrenChanged indicates that a child has been added/removed ( the signal is usually uninteresting as the same information is given by the itemChange hook: QQuickItem::ItemChildAddedChange/QQuickItem::ItemChildRemovedChange ). But it doesn't give any information about geometry or content changes.

examples/blurredbox/MainWindow.cpp
rick-vogel: Thanks for the hint. I activated the clazy check therefore and changed the third argument.

PR #235, opened 2022-11-29
QskGraphic: Don't round before calculating scale factors

### LINKED ISSUE #233: QskGraphic: Wrong scale factor for small painter paths
When drawing a small shape of around 2 pixels width and height, the scale factor doesn't seem to be correct:

QskGraphic-bug

With a bigger shape, more space is used; uncomment the line below to see.

I stripped this down to the following:

`
#include <QtGui>

#include <SkinnyShortcut.h>

#include <QskGraphic.h>
#include <QskGraphicIO.h>
#include <QskGraphicLabel.h>
#include <QskSetup.h>
#include <QskWindow.h>

int main( int argc, char** argv )
{

    SkinnyShortcut::enable( SkinnyShortcut::Quit | SkinnyShortcut::DebugShortcuts );

    QskWindow w;
    w.setPreferredSize( { 500, 400 } );

    QskGraphic graphic;
    QPainter painter( &graphic );

    QPainterPath path1;
    path1.moveTo( 400, 0 );
    path1.addRect( { 400, 0, 50, 50 } );

    QPainterPath path2;
    path2.addRect(0, 0, 2, 2);
    // works:
//    path2.addRect(0, 0, 50, 50);

    painter.drawPath( path1 );
    painter.drawPath( path2 );

    new QskGraphicLabel( graphic, w.contentItem() );

    w.show();
    return app.exec();
}
`

peter-ha:
Otherwise the scale factors might be off considerably.

Resolves #233


peter-ha:
This is the test program:

`
#include <QtGui>

#include <SkinnyShortcut.h>

#include <QskGraphic.h>
#include <QskGraphicIO.h>
#include <QskGraphicLabel.h>
#include <QskSetup.h>
#include <QskWindow.h>

int main( int argc, char** argv )
{

    SkinnyShortcut::enable( SkinnyShortcut::Quit | SkinnyShortcut::DebugShortcuts );

    QskWindow w;
    w.setPreferredSize( { 500, 500 } );

    QskGraphic graphic;
    QPainter painter( &graphic );

    QPainterPath path1;
    path1.moveTo( 400, 0 );
    path1.addRect( { 400, 0, 100, 1 } );

    QPainterPath path2;
    path2.addRect(0, 0, 100, 1);

    painter.drawPath( path1 );
    painter.drawPath( path2 );

    new QskGraphicLabel( graphic, w.contentItem() );

    w.show();
    return app.exec();
}
`

The program produces the following output **without** the patch:

qskgraphic-wrong

The program produces the following output **with** the patch:

qskgraphic-correct

Explanation:

We are trying to paint a path of size (500, 1) into a QskGraphicLabel of size (500, 500).

Now the bounding size of the path is (501, 2) because of the pen size (I guess). This means that we need to scale the size down a little bit to (500, 1.99601) to fit it into the graphic label. 

Since we are operating on integers in `QskGraphicLabel::graphicRect()`, the returned rect from that function gets rounded down to (500, 1).

Now in `QskSkinlet::updateGraphicNode()` we are scaling our graphic default size of (501, 2) into our graphic rect of (500, 1), which results in a scaled size of (250.5, 1) because of the height.

This in turn leads to an X scaling factor of 0.49703 for our painter paths in `QskGraphicPrivate::PathInfo::scaleFactorX()`.

Result: The graphic is rendered way to slow and half of the width is wasted.

uwerat (commented):
The patch removes an optimization, that has an effect when the geometry of the embedded graphic changed below pixel sizes.

F.e. when a label is located inside of a button its position might change due to the animated button shift ( or changing border widths ), when the button is pressed. As animations interpolate between pressed/unpressed this end in small increments.

With this patch we probably end up with creating a new texture for each of these increments ?

peter-ha:
> With this patch we probably end up with creating a new texture for each of these increments ?

I see; is there any alternative to rounding down the graphic rect to full integers? Like qRound() instead of casting to int?
The potential inaccuracy is substantial, as the example shows...

uwerat:
A texture size is always in integers. So the problem of the current implementation is obviously that it does not anticipate the correct one for the given example and IMHO it would be better to fix the wrong anticipation instead of removing it.

peter-ha:
The new commit ensures the texture is big enough by rounding up to full integers.

@uwerat  If you want the fix to be different, could you please point me into the right direction?

peter-ha:
I think the patch is correct:

By rounding up to the next integer when calculating the subcontrol rect, we sort of create a "bounding" rectangle, which will never exceed the contents rect (the subcontrol rect is (500x2) in our case above).

Now when we draw the graphic in QskSkinlet::updateGraphicNode(), we are scaling our graphics bounding rect of (501x2) down into the subcontrol rect of (500x2), which results in a size of (500x1.99601).
So the subcontrol rect seems to behave correctly with the patch.

Btw. with the QskPushButton we use the QskSubControlLayoutEngine; not sure if this is an option for the future for the QskGraphicLabel as well...

PR #238, opened 2022-12-20
IOT dashboard: Fix styling and make more interactive

peter-ha:
@uwerat : One thing I noticed when clicking the buttons in the indoor temperature and humidity sections is that when clicking fast there is no visual feedback of being in the Pressed state.

Is it because the state is left again before the control is redrawn?

And is there a way to ensure the Pressed state is drawn every time the user clicks on it?

### DIFF
--- a/examples/iotdashboard/BoxWithButtons.cpp
+++ b/examples/iotdashboard/BoxWithButtons.cpp
@@ -19,6 +19,8 @@ namespace
 {
     class UpAndDownBox : public QskLinearBox
     {
+        Q_OBJECT
+
       public:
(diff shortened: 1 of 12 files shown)


uwerat:
The animations initiated from a state change are immediately replaced by the animation of a following state change. In case of a button press this leads to the effect, that you are describing. Actually this is a problem, that is known since the very early days of this project.

Could you please create a new issue, so that this does not get lost again ...

peter-ha:
> Could you please create a new issue, so that this does not get lost again ...

Created [issue 240](https://github.com/uwerat/qskinny/issues/240).

Issue #244, opened 2023-01-03
Add skinnable storage page to IOT example


vrcomputing:
![image](https://user-images.githubusercontent.com/6728289/210376626-0dc2b37d-c51b-41b9-8204-6092df54a2a4.png)

- [ ] add list of storages
- [ ] add occupancy bar for e.g. multi media files (music, pictures, ...)
- [ ] add storage capacity / occupancy values ( GiBi or % )

peter-ha:
Has already been merged to master

PR #246, opened 2023-01-05
Add IOT example storage page

vrcomputing:
Solves #244 

iotstoragepage


examples/iotdashboard/CircularProgressBar.cpp
```diff
@@ -74,6 +74,8 @@ class CircularProgressBar::PrivateData
     bool isIndeterminate = false;
 };
 
+CircularProgressBar::~CircularProgressBar() = default;
```
vrcomputing, line 77: Had to implement destructor due to private data unique ptr

PR #247, opened 2023-01-06
Material 3: fix push button size

peter-ha:
Turns out the button was too small, hence the skewed display.

### DIFF
--- a/skins/material3/QskMaterial3Skin.cpp
+++ b/skins/material3/QskMaterial3Skin.cpp
@@ -151,10 +151,10 @@ namespace
         const QskMaterial3Theme& m_pal;
     };
 
-    QFont createFont( int pixelSize, qreal tracking, QFont::Weight weight )
+    QFont createFont( const QString& name, int lineHeight, int size, qreal tracking, QFont::Weight weight )
     {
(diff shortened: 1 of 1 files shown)


peter-ha:
Seems correct comparing to the controls on Figma at 100%:

Screenshot from 2023-01-06 09-37-27

skins/material3/QskMaterial3Skin.cpp
```diff
@@ -151,10 +151,10 @@ namespace
         const QskMaterial3Theme& m_pal;
     };
 
-    QFont createFont( int pixelSize, qreal tracking, QFont::Weight weight )
+    QFont createFont( const QString& name, int lineHeight, int size, qreal tracking, QFont::Weight weight )
     {
-        QFont font( "Roboto" );
-        font.setPixelSize( pixelSize );
+        QFont font( name, size );
```
peter-ha, line 156: not sure if we need to set the point size, since we set the pixel size later anyhow. However, for some reason Material always gives both metrics, see e.g. https://m3.material.io/styles/typography/type-scale-tokens#d74b73c2-ac5d-43c5-93b3-088a2f67723d

PR #251, opened 2023-01-21
push buttons: Support more alignments and style for M3

peter-ha:


peter-ha:
image

src/layouts/QskSubcontrolLayoutEngine.cpp
```diff
@@ -324,7 +323,8 @@ qreal QskSubcontrolLayoutEngine::spacing() const
 
 void QskSubcontrolLayoutEngine::setGraphicTextElements( const QskSkinnable* skinnable,
     QskAspect::Subcontrol textSubcontrol, const QString& text,
-    QskAspect::Subcontrol graphicSubControl, const QSizeF& graphicSize )
+    QskAspect::Subcontrol graphicSubControl, const QSizeF& graphicSize,
+    Qt::Alignment alignment )
```
uwerat, line 327: Qt::Alignment is less expressive than what can configured with setExtraSpacingAt - e.g. the default setting ( -> distributing the space in equal parts left/between/right ) is missing.

So it would be better to pass the alignment as an extra method, so that it is possible to call setGraphicTextElements without killing the default value of extraSpacingAt. As there is already  QskLayoutEngine2D::setDefaultAlignment() with a different meaning it might be something like or setFixedContent( Qt::Orientation, Qt::Alignment ) ?

peter-ha: added a new method as suggested

PR #252, opened 2023-01-25
QskGraphic: Fix scale factors

### LINKED ISSUE #250: QskGraphic: Scale factors are still off for painter paths
The following program calculates zoom factors for painter paths that are still too small for the attached shape (needs to be unpacked first).

Screenshot from 2023-01-18 12-06-55

```cpp
#include <QtGui>

#include <SkinnyShortcut.h>

#include <QskGraphic.h>
#include <QskGraphicIO.h>
#include <QskGraphicLabel.h>
#include <QskSetup.h>
#include <QskWindow.h>

int main( int argc, char** argv )
{

    SkinnyShortcut::enable( SkinnyShortcut::Quit | SkinnyShortcut::DebugShortcuts );

    QskWindow w;
    w.setPreferredSize( { 500, 500 } );

    QskGraphic  = QskGraphicIO::read( QStringLiteral( "shape.graphic" ) );

    new QskGraphicLabel( graphic, w.contentItem() );

    w.show();
    return app.exec();
}
```

[shape.graphic.tar.gz](https://github.com/uwerat/qskinny/files/10445300/shape.graphic.tar.gz)

peter-ha:
We need to know the bounding rect of the graphic to know by how much we can scale our shape, so we need to pass that to the scaling methods.
E.g. when there are two shapes whose pen sizes overlap the path rect left and right, we would calculate wrong scale factors without the grapics' bounding rect.

Resolves #250

### DIFF
--- a/src/graphic/QskGraphic.cpp
+++ b/src/graphic/QskGraphic.cpp
@@ -257,18 +257,20 @@ namespace QskGraphicPrivate
         }
 
         inline double scaleFactorX( const QRectF& pathRect,
-            const QRectF& targetRect, bool scalePens ) const
+            const QRectF& targetRect, const QRectF& graphicBoundingRect, bool scalePens ) const
         {
             if ( pathRect.width() <= 0.0 )
                 return 0.0;
 
             const QPointF p0 = m_pointRect.center();
 
-            const qreal l = qAbs( pathRect.left() - p0.x() );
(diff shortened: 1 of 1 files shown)


peter-ha:
Seems to work now:

image

peter-ha:
Fixes other use cases too:

Old:

triangle-wrong

New:

triangle-correct

PR #255, opened 2023-01-30
Segmented bar: Implement for Material 3

peter-ha:
Screenshot from 2023-01-30 08-35-37

Screenshot from 2023-01-30 08-35-43


peter-ha:
For this to work I had to revert commit fff0389262b2c083fc46e998c5004e031e8dcfc7, which messes up the box rendering

uwerat:
> For this to work I had to revert commit [fff0389](https://github.com/uwerat/qskinny/commit/fff0389262b2c083fc46e998c5004e031e8dcfc7), which messes up the box rendering

The commit should have been correct - the old code was simply wrong. However there are issues with the box renderer, where some of them might have been fixed in in the boxrenderer branch. One of the shortcomings is in combination with irregular radii that did not happen because of the broken implementation of QskBoxShapeMetrics::toAbsolute.

The current state of the implementation in the boxrenderer branch should work 100% correct for all type of linear gradients, that do not require inserting extra gradient lines. So in case the M3 gradients are made of <= 2 colors we could solve your issue by merging the boxrenderer branch into master. 

Also: QskBoxNode creates a QskBoxFillNode instead of a QskBoxRectangleNode in case of QskBoxRenderer::isGradientSupported() returns false. As QskBoxFillNode works with all type of gradients and shapes without any issues I would prefer to exclude broken situations by returning false there - until they have been fixed.

QskBoxRectangleNode is always needed for the border. For the filling QskBoxFillNode is superior - beside of batching. If this wouldn't be somehow important I would throw all difficult parts of QskBoxRenderer and friends away.

peter-ha:
> > For this to work I had to revert commit [fff0389](https://github.com/uwerat/qskinny/commit/fff0389262b2c083fc46e998c5004e031e8dcfc7), which messes up the box rendering
> 
> The commit should have been correct - the old code was simply wrong. However there are issues with the box renderer, where some of them might have been fixed in in the boxrenderer branch. One of the shortcomings is in combination with irregular radii that did not happen because of the broken implementation of QskBoxShapeMetrics::toAbsolute.
> 
> The current state of the implementation in the boxrenderer branch should work 100% correct for all type of linear gradients, that do not require inserting extra gradient lines. So in case the M3 gradients are made of <= 2 colors we could solve your issue by merging the boxrenderer branch into master.

ok, I don't mind either way - We can merge the box renderer branch into master, or I can continue working by keeping the reverted commit locally in my branch. Either way my code doesn't touch any of the box rendering, so we should be fine there...

src/controls/QskSegmentedBar.h
```diff
@@ -48,6 +48,10 @@ class QSK_EXPORT QskSegmentedBar : public QskControl
 
     int addText( const QString& );
     int addGraphic( const QUrl& );
+    int addGraphicAndText( const QUrl&, const QString& );
+
+    QString textAt( int ) const;
+    QskGraphic graphicAt( int ) const;
```
peter-ha, line 54: Since we can have both text and graphic at a certain index, I wonder whether we still need
```
QVariant optionAt( int ) const
```
We could return a QPair if both tex and graphic are set, but that seems a bit weird...

uwerat: Guess I once started with copying the API from QskMenu, where optionAt() returns a QVariantList - at least returning a QVariantList() in QskSegmentedBar::optionAt indicates this. The idea might have been to make optionAt being virtual at some point, so that customized segmented bars ( or menus ) could have more than only text and/or graphic.

As I got distracted with other tasks the API was left in an incomplete - or even messed up - state, as we can see here. In general I would like to have very similar APIs for all controls, that do these "select one of the icon/text pairs"  ( combo boxes and certain list views comes to mind as well ) - maybe even having a common base class providing this API.

peter-ha: sounds good; shall I for the scope of this PR just change the API so it resembles the one from QskMenu, i.e. call the methods "void addOption()" and "QskVariantList optionAt()"?
I also wonder whether e.g. textAt() would be more convenient by just returning an empty string if not set...

uwerat: Yes please go ahead and align the APIs. textAt/graphicAt might be convenient, but should rely on the list you get from optionAt.

peter-ha: ok, I changed the API so that we have `addOption()` and `optionAt()` now, just like in QskMenu. I didn't add `textAt()` or `graphicAt()` yet, since QskMenu doesn't have that either.

We could also make `qskValueAt()` that is used in QskMenuSkinlet public, but maybe this is something for a new PR then...

uwerat:
Actually there are only 2 modes for scaling QskBoxShapeMetrics from relative to absolute size, that make sense The first one keeps the aspect ratio of the radii resulting in a true ellipse for a relative size of 100%. The second one results in rounded rectangles with circular ends.

I used "RoundedRectangle"/"Ellipse" for these modes as I thought that names corresponding to the result is easier to understand than using names describing the calculation ( KeepAspectRatio ). However these names are only correct for relative radii of 100% - otherwise "Ellipse" also results in a rounded rectangle.

So if you have a better idea let me know ...

uwerat:
After using the new mode in the boxes example I disliked the names ( not really corresponding to the visual result ) and changed the ames once again. ScalingMode: Circular/Elliptic

PR #259, opened 2023-02-17
QskSpinBox refactored

rick-vogel:


rick-vogel:
> Did you consider to use QskBoundedValueInput as base class ?

No, simply didn't look at it

FYI: fixed the focus chain in this PR

uwerat:
We could introduce a flag hint ( actually simply a bool ) to tell the skinlet to fit the buttons/text into the shape of the panel - let's call it a compact layout mode. If this mode is set the skinlet would ignore the shape hints for the subcontrols and manually adjusts the corner radii according to the shape of the panel.

Maybe we could even combine this flag with the information given by alignmentHint(Layout) into one enum, that is passed as flag hint ?

src/controls/QskSpinBox.cpp
```diff
@@ -4,356 +4,463 @@
...
+    }
+
+    event->ignore();
 }
 
-void QskSpinBox::mousePressEvent(QMouseEvent *event)
+void QskSpinBox::mousePressEvent( QMouseEvent* const event )
 {
-  m_data->saveMousePosition( -1 * qskMousePosition( event ) );
+    m_data->saveMousePosition( -1 * qskMousePosition( event ) );
```
rick-vogel, line 345: How can we do this more elegantly? Currently -pos => pressed +pos == hovered and pos{} == leave

Maybe somethin like:

```cpp
q->setSkinHint( QskSpinBox::Layout | QskAspect::Flag | QskAspect::Position, Qt::MouseButtonPress );
```

We could also set the subcontrol's state in the QskSpinBox, but this feels wrong to have two locations manipulating states. Currently all/the majority of states are being manipulated by the skinlet.

rick-vogel:
@uwerat I did some major refactoring ( regarding naming and readability ). I hope you didn't change anything in the merged code yet, so consider merging this PR first before taking over.

PR #260, opened 2023-02-17
Add combo box control

peter-ha:
image


peter-ha:
The ripple node role is not implemented for now, because this will only become more important once we have text input in the combo box.

src/controls/QskComboBox.cpp
```diff
@@ -0,0 +1,276 @@
...
+QSK_SYSTEM_STATE( QskComboBox, Pressed, QskAspect::FirstSystemState << 1 )
+QSK_SYSTEM_STATE( QskComboBox, MenuOpen, QskAspect::FirstSystemState << 2 )
+
+class QskComboBox::PrivateData
+{
+  public:
+    PrivateData( QskComboBox* const box )
+        : menu( new QskMenu( box ) )
+    {
+        menu->setPopupFlag( QskPopup::DeleteOnClose, false );
```
uwerat, line 29: menu becomes a dangling pointer after the popup has been closed ( DeleteOnClose ). Guess there will be crashes, when using it a second time ?

src/controls/QskComboBox.cpp
```diff
@@ -0,0 +1,276 @@
...
+    else
+    {
+        hidePopup();
+    }
+
+    setSkinStateFlag( MenuOpen, on );
+    Q_EMIT menuOpenChanged( on );
+}
+
+bool QskComboBox::isMenuOpen() const
```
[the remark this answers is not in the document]
peter-ha: fixed with new patch set

src/controls/QskComboBox.cpp
peter-ha: No, we are setting the flag to false, so it will still be around after closing, right?

PR #279, opened 2023-03-09
virtual keyboard: Make layouts settable

peter-ha:


peter-ha:
Here are screenshots with some keys in the layout rearranged:

Screenshot from 2023-03-09 15-52-58

Screenshot from 2023-03-09 08-30-54

peter-ha:
questions / comments:

* I used variation for styling buttons after all, so the styling is dynamic (instead of creating subclasses): Depending on the language and caps lock mode, some buttons change their variation. This is a lot easier to do when we can set the variation depending on the type of key, compared to creating subclasses. 
* use setMetric() for stretch factor instead of protected method?
* Some of the protected methods could be free functions, but since we want the user to be able to override them, I made them class methods.
* In a next step, we might want to make the static layout with hardcoded locales dynamic. I wonder whether I should make it part of this PR or another one...
* Instead of QskInputPanelBox::setKeyboard, we could also have a virtual createKeyboard() or so. But then we would need to create the layout of the class later than in the constructor because we cannot call the virtual method there.

PR #286, opened 2023-03-14
Add hex (string) literals for QRgb and QColor 

rick-vogel:
# Motivation

When styling in QSkinny often the provided color is in HTML color string, plus this is a subjectively easy and common format. Unfortunately the string needs to be converted at runtime and therefore gets warned by clazy. In order to fix / improve this we could add e.g. a constexpr `""_rgb / ""_rgba` string literal to get the best of both worlds. 

- User must not remember channel reodering from `#RRGGBBAA` to `0xAARRGGBB`
- String literal can still be previewed in Qt Creator ( while hovering the literal with the mouse )
- Optional feature similiar to `std::chrono_literals`
- Optional alpha channel for hex string or integer literals
- User must encode the channel order by using the correct literal

**Proof on Concept:** [Compiler Explorer](https://godbolt.org/z/nWo675rGM)

# Example Usage

```cpp
    #include <QskRgbValue.h>

    {
        using namespace QskRgb::literals;

        const QColor from = "#FF5324"_rgba;
        const QColor to = "#A33517"_rgba;
        setGradient( Q::Fill, { Qt::Vertical, from, to } );
        setGradient( Q::Fill | Q::Pressed, { Qt::Vertical, from.lighter(), to.lighter() } );
        setGradient( Q::Fill | Q::Disabled, { Qt::Vertical, from.darker(), to.darker() } );
    }
```

# Example Usage For QRgb

```cpp
    using namespace QskRgb::literals::qrgb;

    static_assert( "#123456"_rgba == 0xFF123456, "not constexpr" );
    static_assert( "#123456"_argb == 0xFF123456, "not constexpr" );

    static_assert( "#AA112233"_argb == 0xAA112233, "not constexpr" );
    static_assert( "#112233AA"_rgba == 0xAA112233, "not constexpr" );

    static_assert( 0x112233_rgba == 0xFF112233, "not constexpr" );
    static_assert( 0x112233aa_rgba == 0xaa112233, "not constexpr" );

    static_assert( 0x112233_argb == 0xFF112233, "not constexpr" );
    static_assert( 0x112233aa_argb == 0x112233aa, "not constexpr" );
```

# Example Usage For QColor

```cpp
    using namespace QskRgb::literals::color;

    Q_COMPARE( "#112233"_rgba , QColor::fromRgb( 0x11, 0x22, 0x33 ) );
    Q_COMPARE( "#112233AA"_rgba , QColor::fromRgb( 0x11, 0x22, 0x33, 0xAA ) );

    Q_COMPARE( 0x112233_rgba , QColor::fromRgb( 0x11, 0x22, 0x33 ) );
    Q_COMPARE( 0x112233AA_rgba , QColor::fromRgb( 0x11, 0x22, 0x33, 0xAA ) );
```

# References

- https://en.cppreference.com/w/cpp/language/user_literal
- https://doc.qt.io/qt-5/qcolor.html#QColor-3 
- https://en.cppreference.com/w/cpp/chrono/operator%22%22h


src/common/QskRgbLiterals.h
```diff
@@ -0,0 +1,142 @@
...
+
+namespace QskRgb
+{
+    namespace literals
+    {
+        /// @brief converts HTML #RRGGBB[AA] color hex string to Qt's RGB 0xAARRGGBB integer
+        /// @param str The string literal
+        /// @param len The length of the literal
+        /// @note non-hexadecimal characters are replaced with 0x0 value
+        constexpr unsigned int operator""_rgba( const char* const str, const size_t len )
```
rick-vogel, line 17: Should / could be Q_REQUIRED_RESULT + Q_DECL_CONSTEXPR

src/common/QskRgbLiterals.h
```diff
@@ -0,0 +1,142 @@
+/******************************************************************************
+ * Copyright (C) 2023 Edelhirsch Software GmbH
+ * This file may be used under the terms of the 3-clause BSD License
+ *****************************************************************************/
+
+#ifndef QSK_RGB_LITERALS_H
+#define QSK_RGB_LITERALS_H
+
```
rick-vogel, line 8:
```cpp
#include <cstddef>
```

src/common/QskRgbLiterals.h
```diff
@@ -0,0 +1,142 @@
...
+namespace QskRgb
+{
+    namespace literals
+    {
+        /// @brief converts HTML #RRGGBB[AA] color hex string to Qt's RGB 0xAARRGGBB integer
```
uwerat, line 13: Please write the comments as plain text - the doxygen parser will never see this file and using its syntax makes the documentation less readable.

The doxygen generated documentation will be generated from files in the doc folder only - and yes this approach is against the basic idea of having implementation and documentation in the same place.

uwerat (changes requested):

src/common/QskRgbLiterals.h
uwerat: Q_DECL_CONSTEXPR was a placeholder to make the code compilable without c++11.

Q_REQUIRED_RESULT is still valid - but when introducing these type of attributes it should be done throughout the complete qskinny code base. Maybe later ...

rick-vogel: Do we (or Qt) define any requirement towards a minimum C++ standard version?

src/common/QskRgbValue.h
```diff
@@ -235,4 +235,136 @@ namespace QskRgb
...
+        {
+            return 0;
+        }
+
+        if ( str[ 0 ] != '#' )
+        {
+            return 0;
+        }
+
+        QRgb rgb = 0x00;
```
rick-vogel, line 254:
```
In file included from common/QskRgbValue.cpp:6:
common/QskRgbValue.h:255:14: warning: variable declaration in a constexpr function is a C++14 extension [-Wc++14-extensions]
        QRgb rgb = 0x00;
             ^
```

uwerat: see #287 - will do this today.

uwerat: done: using c++17 features is allowed now

src/common/QskRgbValue.h
```diff
@@ -235,4 +235,150 @@ namespace QskRgb
...
+            const char* const str, const size_t len ) noexcept
+        {
+          const auto argb = static_cast<int>(fromHexString( str, len ));
+          const auto r = ( argb >> 16) & 0xFF;
+          const auto g = ( argb >> 8) & 0xFF;
+          const auto b = ( argb >> 0) & 0xFF;
+          const auto a = ( argb >> 24) & 0xFF;
+          const auto color = QColor{r,g,b,a};
+          return color;
+        }
```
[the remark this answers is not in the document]
rick-vogel:
```cpp
    {
        using namespace QskRgb::literals::qrgb;
        Q_ASSERT( "#112233AA"_rgba == 0xAA112233 );
        Q_ASSERT( "#112233AA"_argb == 0x112233AA );
    }

    {
        using namespace QskRgb::literals::color;
        Q_ASSERT( "#112233AA"_rgba == QColor::fromRgb( 0x11, 0x22, 0x33, 0xAA ) );
        Q_ASSERT( "#112233AA"_argb == QColor::fromRgb( 0x22, 0x33, 0xAA, 0x11 ) );
    }
```

src/common/QskRgbLiterals.h
uwerat: Qt6: C++17, since yesterday QSkinny has the same. So it is allowed to rely on C++17

src/common/QskRgbValue.h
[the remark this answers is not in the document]
uwerat: QColor is very present in the Qt API, so that it needs to be supported in the public API of QSkinny, However I'm thinking since quite some time of replacing it by QRgb in many places especially in QskGradientStop. Actually I'm even tempted to introduce our own color class - something like this:

```
class QskColor
{
    ...
    private:
        QRgb m_rgba;
};
```

This way it would be possible to have automatic conversions from QRgb/QColor/Qt::GlobalColor. It would also avoid the painful "QColor(QRgb rgb)" trap, where the alpha value gets lost - usually by accident.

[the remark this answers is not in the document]
rick-vogel: I mean now we have both ""_rgba/argb -> QRgb + QColor, and the user can/must use the suffix that defines the channel order in the string literal. One more measure to make to user think about it. Additionally we can now profit from the QColor constexpr C-TOR, so now we really have the flexibility and best of both world.

src/common/QskRgbValue.h
```diff
@@ -235,4 +235,335 @@ namespace QskRgb
...
+                const auto g = ( argb >> 8 ) & 0xFF;
+                const auto b = ( argb >> 0 ) & 0xFF;
+                const auto a = ( len == max ? argb >> 24 : 0xFF ) & 0xFF;
+                return qRgba( r, g, b, a );
+            }
+
+            // converts a hex literal from '0xRRGGBB[AA]' to '0xAARRGGBB' integer
+            template< char... Cs >
+            constexpr QRgb operator""_rgba()
+            {
```
rick-vogel, line 409: @uwerat I had to resort to this `numeric literal operator template` since AFAIK there are no 32 bit literal operators (see: https://en.cppreference.com/w/cpp/language/user_literal). Thereby a user can type `0x112233_argb` or `0x112233_argb` and we can detect and default the missing alpha channel.

src/common/QskRgbValue.h
```diff
@@ -235,4 +235,335 @@ namespace QskRgb
...
+                    const auto r = ( rgba >> ( 4 * 4 ) ) & 0xFF;
+                    const auto g = ( rgba >> ( 2 * 4 ) ) & 0xFF;
+                    const auto b = ( rgba >> ( 0 * 4 ) ) & 0xFF;
+                    const auto a = ( rgba >> ( 6 * 4 ) ) & 0xFF;
+                    return qRgba( r, g, b, a );
+                }
+            }
+
+#define QSK_REQUIRE_CONSTEXPR_LITERAL 1
+#ifdef QSK_REQUIRE_CONSTEXPR_LITERAL
```
rick-vogel, line 461: @uwerat how do we want to proceed with this? I think it can stay and the user can decide to add `DEFINES += QSK_REQUIRE_CONSTEXPR_LITERAL` in order to guarantee `constexpr` literals.

Obviously the line above must be removed before merging: `#define QSK_REQUIRE_CONSTEXPR_LITERAL 1`

uwerat: Why not moving these assertions to QskRgbValue.cpp. It should be good enough to do them once - when building the qskinny lib.

But in the end I do not believe, that we need to keep these assertions at all. It is not very likely that someone will touch the code after it is working once and in case it is necessary it is usually because of something that has not been found by those checks.

src/common/QskRgbValue.h
```diff
@@ -235,4 +235,335 @@ namespace QskRgb
...
+
+            // ARGB hex literal to QRgb with defaulted alpha = 0xFF
+            static_assert( 0x112233_argb == 0xFF112233, "" );
+            static_assert( 0xaabbcc_argb == 0xFFAABBCC, "" );
+            static_assert( 0xAABBCC_argb == 0xFFAABBCC, "" );
+            static_assert( 0x112233aa_argb == 0x112233aa, "" );
+#endif
+        }
+
+        namespace color
```
rick-vogel, line 484: @uwerat rename to `namespace qcolor` to be consistent with `namespace qrgb`?

uwerat: Please no qrgb and no qcolor - even for nested namespaces. 'Q' or 'q' names are reserved for Qt classes.

Not sure if the "literals" namespace is necessary as the definitions are already inside of QskRgb. I guess it would also be better to use different literals ( or function names ) instead of the additioal rgb/color namespace nesting. Otherwise users always have to qualify each literal when using QColor ad QRgb in the same file.

src/common/QskRgbValue.h
rick-vogel: The idea was to use the assertion both as constexpr check and documentation / example, but it's fine for me to move it into the cpp. Although when thinking about prebuilt packaged the user would have to write his own checks...

The whole static assertion is just to guarantee that the literals are constexpr since the keyword itself doesn't guarantee it.

uwerat: Then it would be good enough to have only one static assertion. But in case there is is some compiler that fails, what is the user supposed to do with this information ?

IMHO the checks are not worth to introduce a compile time penalty to each file that is including the literals.

tests/tests_main.cpp
```diff
@@ -0,0 +1,4 @@
+#include "tests_main.h"
+#include <QTest>
+
+QTEST_APPLESS_MAIN( QskRgbLiterals )
```
uwerat, line 4: g++ --version
g++ (Debian 10.2.1-6) 10.2.1 20210110
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

g++ -c -pipe -isystem /opt/qskinny/config/qsk3/Qt/include/QtCore -isystem /opt/qskinny/config/qsk3/Qt/include/QtCore/6.4.2/QtCore -isystem /opt/qskinny/config/qsk3/Qt/include/QtGui -isystem /opt/qskinny/config/qsk3/Qt/include/QtGui/6.4.2/QtGui -isystem /opt/qskinny/config/qsk3/Qt/include/QtQuick -isystem /opt/qskinny/config/qsk3/Qt/include/QtQuick/6.4.2/QtQuick -isystem /opt/qskinny/config/qsk3/Qt/include/QtQml -isystem /opt/qskinny/config/qsk3/Qt/include/QtQml/6.4.2/QtQml -pedantic-errors -Wextra -Werror=format-security -Wsuggest-override -Wno-deprecated-copy -Wno-deprecated-declarations -g -O0 -std=c++1z -fvisibility=hidden -fvisibility-inlines-hidden -Wall -Wextra -fPIC -D_REENTRANT -DQT_STRICT_ITERATORS -DITEM_STATISTICS=1 -DQT_NO_KEYWORDS -DQT_QUICK_LIB -DQT_OPENGL_LIB -DQT_GUI_LIB -DQT_TESTLIB_LIB -DQT_QMLMODELS_LIB -DQT_QML_LIB -DQT_QMLINTEGRATION_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -DQT_TESTCASE_BUILDDIR='"/opt/qskinny/build/qsk3/tests"' -I/opt/qskinny/qsk3/tests -I. -I/opt/qskinny/qsk3/src/common -I/opt/qskinny/qsk3/src/controls -I/opt/qskinny/qsk3/src/dialogs -I/opt/qskinny/qsk3/src/graphic -I/opt/qskinny/qsk3/src/inputpanel -I/opt/qskinny/qsk3/src/layouts -I/opt/qskinny/qsk3/src/nodes -I/opt/qskinny/config/qsk3/Qt/include -I/opt/qskinny/config/qsk3/Qt/include/QtQuick -I/opt/qskinny/config/qsk3/Qt/include/QtOpenGL -I/opt/qskinny/config/qsk3/Qt/include/QtGui -I/opt/qskinny/config/qsk3/Qt/include/QtTest -I/opt/qskinny/config/qsk3/Qt/include/QtQmlModels -I/opt/qskinny/config/qsk3/Qt/include/QtQml -I/opt/qskinny/config/qsk3/Qt/include/QtQmlIntegration -I/opt/qskinny/config/qsk3/Qt/include/QtNetwork -I/opt/qskinny/config/qsk3/Qt/include/QtCore -Imoc -I/opt/qskinny/config/qsk3/Qt/mkspecs/linux-g++ -o obj/tests_main.o /opt/qskinny/qsk3/tests/tests_main.cpp
/opt/qskinny/qsk3/tests/tests_main.cpp:4:36: error: ISO C++11 requires at least one argument for the "..." in a variadic macro
    4 | QTEST_APPLESS_MAIN( QskRgbLiterals )
      |                                    ^
make: *** [Makefile:466: obj/tests_main.o] Error 1

src/common/QskRgbValue.h
```diff
@@ -236,4 +236,312 @@ namespace QskRgb
...
+            return parse_hex_literal_h< 0, Cs... >();
+        }
+
+        template< char C, char... Cs >
+        constexpr QRgb parse_hex_literal_0()
+        {
+            static_assert( C == '0', "invalid hex prefix character" );
+            return parse_hex_literal_x< Cs... >();
+        }
+
```
uwerat, line 378: Why not ( there is probably a better name than "fromHexRgba" ):

```
        constexpr QRgb fromHexRgba(
            const char* const str, const size_t len ) noexcept
        {
            constexpr auto max = 9;
            const auto argb = static_cast< int >( fromHexString( str, len ) );
            const auto r = ( argb >> ( len == max ? 24 : 16 ) ) & 0xFF;
            const auto g = ( argb >> ( len == max ? 16 : 8 ) ) & 0xFF;
            const auto b = ( argb >> ( len == max ? 8 : 0 ) ) & 0xFF;
            const auto a = ( len == max ? argb & 0xFF : 0xFF );
            return qRgba( r, g, b, a );
        }

```
and then

```
        namespace integral
        {
            QSK_EXPORT constexpr QRgb operator""_rgba(
                const char* const str, const size_t len ) noexcept
            {
                return fromHexRgba( str, len );
            }
        }

        namespace color
        {
            QSK_EXPORT constexpr QColor operator""_rgba(
                const char* const str, const size_t len ) noexcept
            {
                const auto c = fromHexRgba( str, len );
                return { qRed( c ), qGreen( c ), qBlue( c ), qAlpha( c ) };
            }
    }
```
Guess it is possible to add a 3rd parameter - indicating the order - to fromHexString, so that we do not need fromHexRgba anymore:

```
           QSK_EXPORT constexpr QRgb operator""_rgba(
                const char* const str, const size_t len ) noexcept
            {
                return fromHexString( str, len, RGBA );
            }

           QSK_EXPORT constexpr QRgb operator""_argb(
                const char* const str, const size_t len ) noexcept
           {
                return fromHexString( str, len, ARGB );
           }
```

PR #289, opened 2023-04-03
color filter: Make sure to detach from original brush when substituting

peter-ha:
QBrush does not behave like a normal value class, but is explicitly shared, so we need to create a new brush and set the new gradients when replacing them.

@uwerat : I added another commit for an auto test, if you want I can only commit the code for now. But I thought it might be helpful not to lose the code to verify that it is working correctly; let me know what you think...

For the auto tests I tried to keep the same structure as Qt has with their tests, i.e. one directory for each class, so you can run tests (or "make check") for all classes (nice for CI) but also for individual classes.


peter-ha:
Reverted the auto test for now

uwerat:
merged into master

PR #300, opened 2023-04-10
Adjust generated target's folder property

vrcomputing:
The generated Qt `*_qmlimportscan` targets clutter the IDE and should therefore be either put next to their base target's folder or grouped / hidden aways in a collective folder.

Before

image

After

image

### DIFF
--- a/cmake/QskConfigMacros.cmake
+++ b/cmake/QskConfigMacros.cmake
@@ -163,3 +163,23 @@ macro(qsk_finalize_build_flags)
     endif()
 
 endmacro()
+
+function(qsk_internal_sync_generated_target_folder directory)
+    # loop through all targets in the directory recursively and
+    # synchronize the generated target's folder property with 
+    # the actual target's folder property
+    get_property(TARGETS DIRECTORY "${directory}" PROPERTY BUILDSYSTEM_TARGETS)
+    foreach(target IN LISTS TARGETS)
+        if(TARGET ${target}_qmlimportscan)
+            get_target_property(folder ${target} FOLDER)
+            set_target_properties(${target}_qmlimportscan PROPERTIES FOLDER ${folder})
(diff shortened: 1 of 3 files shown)


cmake/QskConfigMacros.cmake
```diff
@@ -163,3 +163,23 @@ macro(qsk_finalize_build_flags)
...
+
+function(qsk_internal_sync_generated_target_folder directory)
+    # loop through all targets in the directory recursively and
+    # synchronize the generated target's folder property with 
+    # the actual target's folder property
+    get_property(TARGETS DIRECTORY "${directory}" PROPERTY BUILDSYSTEM_TARGETS)
+    foreach(target IN LISTS TARGETS)
+        if(TARGET ${target}_qmlimportscan)
+            get_target_property(folder ${target} FOLDER)
+            set_target_properties(${target}_qmlimportscan PROPERTIES FOLDER ${folder})
```
vrcomputing, line 175: @uwerat we could also define a collective "generated" "folder" for the IDE's to hide these implementation / Qt / CMake specific targets e.g.:

image

vrcomputing: PS: should we put the `qskqmlexport` + `qsktestsupport` targets into the `libs` folder aswell?

uwerat:
Actually we do not want to run qmlimportscanner at all and I would prefer to avoid these calls instead of postprocessing an irrelevant result. Unfortunately I couldn't find an easy way to suppress these calls once we have a dependency to Qt6Qml that is implicitly set from Qt6QuickDependencies.cmake.

Possible workarounds for development environments:

- using add_library instead of qt6_add_library
- using MANUAL_FINALIZATION for qt6_add_library without calling qt6_finalize_target afterwards

uwerat:
Kudos to Rick: qmlimportscanner can be blocked by setting the target property QT_QML_MODULE_NO_IMPORT_SCAN.
Then the patch is obsolete as we do not generate those qmlimportscan directories anymore.

PR #317, opened 2023-06-01
keyboard: Differentiate between pressed and clicked

peter-ha:
For pressed, we want to make use of the auto repeat feature, in contrast to the clicked case.

### DIFF
--- a/src/inputpanel/QskVirtualKeyboard.cpp
+++ b/src/inputpanel/QskVirtualKeyboard.cpp
@@ -383,6 +383,9 @@ void QskVirtualKeyboard::ensureButtons()
                 connect( button, &QskPushButton::pressed,
                     this, &QskVirtualKeyboard::buttonPressed );
 
+                connect( button, &QskPushButton::clicked,
+                    this, &QskVirtualKeyboard::buttonClicked );
+
                 m_data->keyButtons += button;
             }
         }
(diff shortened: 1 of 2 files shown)


uwerat:
From reading the code I would expect that the only effect of these changes is to process the keys Return/Enter/Shift/CapsLock/ModeWitch on release instead of press - right ?

But what is the intention behind this ?

peter-ha:
> From reading the code I would expect that the only effect of these changes is to process the keys Return/Enter/Shift/CapsLock/ModeWitch on release instead of press - right ?

yes, that is correct.

> 
> But what is the intention behind this ?

The intention is to make the code more robust: If we reacted on pressed for the special keys that trigger e.g. accept() or reject(), we emit a signal which might lead to the keyboard being destroyed. Now after pressed there would also be a released signal, which might try to access buttons etc. that are already destroyed.
Of course we have the "if(button == nullptr)" check in there anyhow, but it still seems strange to try to send events to nonexistant items.

Apparently there is also an issue when opening the keyboard from a recursive event loop; I can ask more details about the case if you want.

Also, isn't it the normal thing to react to clicked instead of pressed? With the normal keyboard buttons we have the special case that we want to use the auto repeat feature, hence we react on pressed in that case; for the special keys we react on clicked like we do with other buttons.

peter-ha:
Update: I checked various keyboards and documented my findings here: https://gist.github.com/peter-ha/469cc8bf27cf39bb211eff5ba83f399e

Summary: The Qt Virtual Keyboard reacts only on release (i.e. clicked), because they have a context menu pop up for most keys with tap and hold. Ubuntu and Windows react on press for normal keys and release for special keys, Mac reacts only on pressed.

IMHO what makes most sense for QSkinny is to envision a context menu for the future like QtVKB (and also Android) has; this would mean no autorepeat for normal keys, and for consistency we would enter the key on release (right?). It seems we would still need to let customers opt-in for auto-repeat, and in that case we have to enter the key on press. I.e. something like:

```
if( isAutoRepeat(key)) {
    handleKeyInPressed();
} else {
    handleKeyInReleased(); // or tap and hold for context menus
}

uwerat:
Is there anything we can deduce from: https://doc.qt.io/qt-6/qstylehints.html ?

PR #319, opened 2023-06-12
segmented bar: Allow subclassing

peter-ha:


uwerat:
The toplevel layout classes should indeed be available for application code, but I would like to check if there aren't parts that should stay internal ( f.e QskLayoutChain/QskLayoutElement ) first.

However I'm hesitating with the current implementation of QskSubcontrolLayoutEngine. I implemented the initial version of this class to avoid having to duplicate the layout code for text/graphic combinations in various controls - having in mind, that this implementation will be for internal use only.

But using the implemented layout algorithms ( linear + grid ) for arranging subcontrols makes a lot of sense - we might be able to replace a lot more layout code in the skinlets than horizontal/vertical text/graphic combinations only. This needs more work to be done on QskSubcontrolLayoutEngine with maybe substantial API changes. That's why I would prefer not to export the API too early.

peter-ha:
> But using the implemented layout algorithms ( linear + grid ) for arranging subcontrols makes a lot of sense - we might be able to replace a lot more layout code in the skinlets than horizontal/vertical text/graphic combinations only. This needs more work to be done on QskSubcontrolLayoutEngine with maybe substantial API changes. That's why I would prefer not to export the API too early.

Fair enough - Shall I remove the commit which exports the layout engines, and still keep the part which allows for overriding the segment size of the segmented bar skinlet? That latter part is still valid I guess (and the class is exported anyhow)...

uwerat:
> Fair enough - Shall I remove the commit which exports the layout engines, and still keep the part which allows for overriding the segment size of the segmented bar skinlet? That latter part is still valid I guess (and the class is exported anyhow)...

The current implementation internally assumes, that all segments have the same size. With the patch this assumption becomes part of the API - not allowing to have different segment sizes.

I'm thinking about a more general solution - something like:

    - virtual QskSkinlet::sampleSizeHint( ..., QskAspect::Subcontrol, int index );

peter-ha:
> I'm thinking about a more general solution - something like:
> 
> ```
> - virtual QskSkinlet::sampleSizeHint( ..., QskAspect::Subcontrol, int index );
> ```

I see, this is indeed more flexible. I'll have to see whether I can do this soon at work; but shouldn't be too much work anyhow...

PR #329, opened 2023-07-05
skins: Switch automatically when color scheme changes

peter-ha:
https://github.com/uwerat/qskinny/assets/1749504/765c65f6-e8ba-42ea-b805-ed02e6acbfdb


peter-ha:
Let me know what you think of the idea. We could also make QskSkin::SkinInfo a class of its own...

uwerat:
With this patch a skin is identified by the factory instance and the tuple name/scheme, that has to be unique inside of the factory. However the names of the provided skins factories are the same withing each factory and the name is actually the name of the factory. So it would be more accurate to have something like this:
```
class QskSkinFactory
{
   ...
    QString name() const;
    QVector< QskSkin::ColorScheme > colorSchemes() const; // could be a bitmask as well
}
```
However this limits the concept of a skin factory to always produce skins that only differ by the color scheme. It also makes it impossible to have more than 3 color schemes, what is not unlikely as QskSkin::UnknownScheme is a placeholder for all sort of schemes.

PR #330, opened 2023-07-10
Bugfix: Initialize colors with default black

rick-vogel:


skins/fluent2/QskFluent2Skin.cpp
```diff
@@ -1587,7 +1587,7 @@ void Editor::setupTextInputColors(
 
     for( const auto state : { A::NoState, Q::Hovered, Q::Focused, Q::Editing, Q::Disabled } )
     {
-        QRgb panelColor, borderColor1, borderColor2, textColor;
+        QRgb panelColor = Qt::black, borderColor1 = Qt::black, borderColor2 = Qt::black, textColor = Qt::black;
 
         if ( state == A::NoState )
```
rick-vogel, line 1592: `state` was 0x4000 in my scenario, don't know how and why but nonetheless the variables should be initialized

peter-ha: I think we should rather convert the "if( state == A::NoState)" clause to an else clause that catches all other cases.

Qt::black doesn't work here because we should always use a color from the palette due to the day/night themes (black couldn't be read in the night theme).

peter-ha (changes requested):

peter-ha (approved):
Uwe will have the final word

peter-ha (changes requested):
ah wait I guess we can still get rid of the Qt::black change in line 1590

skins/fluent2/QskFluent2Skin.cpp
rick-vogel: done

peter-ha (approved):

uwerat:
Looks like a copy/paste bug to me: Disabled is initialized twice, while Hovered is missing. Putting NoStyle as default path at the end of the if/else cascade initializes the Hovered state, but not with the correct settings.

Hope this fix ( https://github.com/uwerat/qskinny/commit/2ec50ad238f3d6166fa6a990cfe9edef80e53497 ) does the right thing ?

PR #332, opened 2023-07-17
Style improvements

peter-ha:
The full diff of this PR is probably hard to understand; maybe it is easier looking at each commit individually

### DIFF
--- a/src/controls/QskListView.h
+++ b/src/controls/QskListView.h
@@ -75,6 +75,8 @@ class QSK_EXPORT QskListView : public QskScrollView
     virtual QskColorFilter graphicFilterAt( int row, int col ) const;
 #endif
 
+    QRectF focusIndicatorRect() const override;
+
   public Q_SLOTS:
     void setSelectedRow( int row );
 
--- a/src/controls/QskListView.h
+++ b/src/controls/QskListView.h
@@ -33,7 +33,7 @@ class QSK_EXPORT QskListView : public QskScrollView
     using Inherited = QskScrollView;
 
   public:
-    QSK_SUBCONTROLS( Cell, Text )
+    QSK_SUBCONTROLS( Viewport, Cell, Text )
     QSK_STATES( Selected )
 
     enum SelectionMode
--- a/skins/fluent2/QskFluent2Skin.cpp
+++ b/skins/fluent2/QskFluent2Skin.cpp
@@ -1095,7 +1095,18 @@ void Editor::setupScrollViewMetrics()
     using Q = QskScrollView;
 
     for ( auto subControl : { Q::HorizontalScrollBar, Q::VerticalScrollBar } )
+    {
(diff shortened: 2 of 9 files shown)


src/controls/QskListView.h
```diff
@@ -33,7 +33,7 @@ class QSK_EXPORT QskListView : public QskScrollView
     using Inherited = QskScrollView;
 
   public:
-    QSK_SUBCONTROLS( Cell, Text )
+    QSK_SUBCONTROLS( Viewport, Cell, Text )
```
uwerat, line 36: Is it really necessary to introduce the Viewport subcontrol ?

There are only 2 classes derived from QskScrollView:

- QskListView
- QskScrollArea

I would be surprised if M3 or F2 Style guides have something specified for the background of the scroll area viewport.

If there is none - why not using the settings you have added for the listview as settings for QskScrollView::Viewport ?

But in case there is something specified - is the background of the list really a replacement for the scrollview viewport. F.e when having a list, that is shorter than the viewport we might want to see background an viewport ?

peter-ha: To me it seems that lists have a special background according to https://www.figma.com/file/O4H724CKmUVPocw6JoSUrd/Material-3-Design-Kit-(Community)?type=design&node-id=51964-62981&mode=design&t=J356dDaAjhwxnYxi-0 , which according to Figma is set to surfaces/[light|dark]/surface1.
Since there doesn't seem to be any setting for scroll views (there doesn't seem to be a spec for scrollable content), the background for the scroll view is set to background.
However, the flutter example at https://flutter.github.io/samples/web/material_3_demo/#/ sets all backgrounds to the surface1 color, so we could define that one as the generic background color. This would mean that all pages, cards etc. had that background color though.

uwerat: From looking at https://m3.material.io/components/lists/specs it might be, that the special background is actually the background of the row. At least the picture on top of that page seems to indicate this: the background ( of the viewport ? ) can be seen between the rows - almost like separators.

The current implementation of QskListViewSkinlet creates the cell background nodes only when alternatingRowColors is enabled. But we could change this boolean to an enum with 3 modes:

- Never
- Alternating
- Always

For Mat3 we would have a default setting of Always ?

uwerat: When looking at the specs we can see a divider: https://m3.material.io/components/lists/specs. It seems to be an element of the item/row - not of the container ( = list view ). It is shorter than the width of the item - obviously affected by the same padding as the other content.

But the image on top of https://m3.material.io/components/lists/overview does not have this padding and the color of the "divider" seems to be different than the one on the specs page. Maybe this one was made with spacings between the items and no divider.

Very confusing - IMHO not well done as a specification.

uwerat:
All relevant code is in master now

PR #333, opened 2023-07-20
Add circular coordinate system display

rick-vogel:
image

- provide fine granularity or rotating and hiding individual subcontrols


src/controls/LevelingSensor/QskLevelingSensor.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <unordered_map>
+
+class QSK_EXPORT QskLevelingSensor : public QskControl
+{
+    Q_OBJECT
+        using Inherited = QskControl;
+public:
+    QSK_SUBCONTROLS(
+        OuterDisk,
```
rick-vogel, line 16: maybe `BackgroundDisk` or `BackgroundCircle`

src/controls/LevelingSensor/QskLevelingSensor.h
```diff
@@ -0,0 +1,55 @@
...
+
+public:
+    Q_REQUIRED_RESULT const Tickmarks& tickmarks(Qt::Axis axis) const;
+    Q_REQUIRED_RESULT const TickmarksLabels& tickmarkLabels(Qt::Axis axis) const;
+    Q_REQUIRED_RESULT const QVector3D& angle() const noexcept;
+    Q_REQUIRED_RESULT const QVector3D& subControlRotation(QskAspect::Subcontrol subControl) const noexcept;
+
+private:
+
+    // TODO use pimpl
```
rick-vogel, line 47: - [ ] use pimpl

src/controls/LevelingSensor/QskLevelingSensor.h
```diff
@@ -0,0 +1,55 @@
...
+    void setTickmarks(Qt::Axis axis, Tickmarks tickmarks);
+    void setTickmarksLabels(Qt::Axis axis, TickmarksLabels labels);
+    void setAngle(const QVector3D& degree);
+    void setAngle(Qt::Axis axis, float degree);
+
+    void setSubControlRotation(QskAspect::Subcontrol subControl, const QVector3D& degree);
+
+Q_SIGNALS:
+    void anglesChanged(const QVector3D& degree);
+    void subControlRotationChanged(QskAspect::Subcontrol subControl, const QVector3D& degree);
```
rick-vogel, line 37: maybe `virtual` as well as `virtual`getter

src/controls/LevelingSensor/QskLevelingSensorSkinlet.cpp
```diff
@@ -0,0 +1,461 @@
...
+    const auto shape = QskBoxShapeMetrics{ state.r1 };
+    const auto metrics = sensor->boxBorderMetricsHint( subControl );
+    const auto colors = sensor->boxBorderColorsHint( subControl );
+
+    auto gradient = sensor->gradientHint( Q::Horizon );
+    gradient.setDirection( QskGradient::Linear );
+    gradient.setLinearDirection( Qt::Vertical );
+    gradient.setStops( { { 0.0, gradient.startColor() }, { pY, gradient.startColor() },
+        { pY, gradient.endColor() }, { 1.0, gradient.endColor() } } );
+
```
rick-vogel, line 232: Creating a gradient with stops at 0%, 50%, 50% and 100% to achieve hard line

uwerat: When not having a stop at 0.0 the renderer takes the color of the first step and fills from 0.0 to the first position. Similar, when you do not have a stop at 1.0. So in this particular case it would be enough to have the 2 stops at pY.

src/controls/LevelingSensor/QskLevelingSensorSkinlet.cpp
```diff
@@ -0,0 +1,461 @@
...
+    const auto scale = sensor->strutSizeHint( subControl );
+    const auto r3 = state.r1 * scale.height();
+
+    auto* const clipping =
+        ensureNode< PolygonClipNode, QSGTransformNode, LinearTickmarksNode >(
+            node );
+    auto* const transform = static_cast< QSGTransformNode* >( clipping->firstChild() );
+    auto* const tickmarks = static_cast< LinearTickmarksNode* >( transform->firstChild() );
+
+    auto size = qvariant_cast< QVector3D >( sensor->effectiveSkinHint( subControl ) ) * r3;
```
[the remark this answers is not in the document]
uwerat: My fist idea would be to use QskAspect::Variation, where you have Tiny/Small/Large/Huge. However this would block using other variations like Vertical/Horizontal. So maybe it is not a bad idea to introduce a class for a tuple of increasing sizes - in the best case something that works for other situations as well. 

Of course QVector3D works out of the box, but the QVectorXD classes are float ( not qreal ) based and have primarily been made for OpenGL APIs.

playground/levelingsensor/Dashboard.cpp
```diff
@@ -0,0 +1,125 @@
...
+            labels.reserve( labelsCount );
+
+            for ( int i = 0; i < labelsCount; i++ )
+                labels += QString::number( i );
+
+            setTickLabels( labels );
+        }
+    };
+
+    class SpeedoMeter : public DashboardDial
```
peter-ha, line 61: Speedometer

src/controls/LevelingSensor/QskLevelingSensor.h
```diff
@@ -0,0 +1,45 @@
...
+    using Tickmarks = QskScaleTickmarks;
+    using TickmarksLabels = QVector< QPair< qreal, QString > >;
+
+    explicit QskLevelingSensor( QQuickItem* parent = nullptr );
+    ~QskLevelingSensor();
+
+  public Q_SLOTS:
+    void setTickmarks( Qt::Axis axis, Tickmarks tickmarks );
+    void setTickmarksLabels( Qt::Axis axis, TickmarksLabels labels );
+    void setAngle( const QVector3D& degree );
```
peter-ha, line 26: maybe call the argument degrees or better angles?

src/controls/LevelingSensor/QskLevelingSensor.h
```diff
@@ -0,0 +1,45 @@
...
+    void setAngle( Qt::Axis axis, float degree );
+
+    void setSubControlRotation( QskAspect::Subcontrol subControl, const QVector3D& degree );
+
+  Q_SIGNALS:
+    void anglesChanged( const QVector3D& degree );
+    void subControlRotationChanged( QskAspect::Subcontrol subControl, const QVector3D& degree );
+
+  public:
+    Q_REQUIRED_RESULT const Tickmarks& tickmarks( Qt::Axis axis ) const;
```
peter-ha, line 36: Maybe just return the Tickmarks by value instead of const-ref, which is fine for Qt-style ref-counted classes?

src/controls/LevelingSensor/QskLevelingSensor.cpp
```diff
@@ -0,0 +1,175 @@
...
+        if ( dst != src )
+        {
+            dst = src;
+            return true;
+        }
+        return false;
+    }
+
+    template<>
+    bool compareExchange< float >( float& dst, const float& src )
```
peter-ha, line 23: Does this cover the double case as well?

src/controls/LevelingSensor/QskLevelingSensor.h
peter-ha, line 1: I guess this should go just into controls? And the nodes into src/nodes

src/controls/LevelingSensor/QskLevelingSensorNodes.h
```diff
@@ -0,0 +1,303 @@
...
+
+#include <QFontMetricsF>
+#include <qmath.h>
+
+class RadialTickmarksNode final : public QSGGeometryNode
+{
+public:
+    RadialTickmarksNode() : m_geometry(QSGGeometry::defaultAttributes_Point2D(), 0)
+    {
+        m_geometry.setDrawingMode(QSGGeometry::DrawLines);
```
peter-ha, line 17: some style issues here

src/controls/LevelingSensor/QskLevelingSensorNodes.h
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    QSGGeometry m_geometry;
+    QSGFlatColorMaterial m_material;
+    qreal m_r1 = 0.0;
+    QVector3D m_r2 = { 0.5, 0.75, 1.0 };
+    QskHashValue m_tickmarksHash{ 0 };
+};
+
+class LinearTickmarksNode final : public QSGGeometryNode
```
peter-ha, line 91: use QskTickmarksNode here

src/controls/LevelingSensor/QskLevelingSensorNodes.h
```diff
@@ -0,0 +1,303 @@
...
+    using QSGClipNode::setClipRect;
+
+    QSGGeometry m_geometry;
+    qreal m_radius = 1.0;
+    qreal m_cx = 0;
+    qreal m_cy = 0;
+    int m_count = 360;
+};
+
+template<typename CRTP>
```
peter-ha, line 235: What does CRTP mean? Doesn't sound easy to understand...

src/controls/LevelingSensor/QskLevelingSensorSkinlet.cpp
```diff
@@ -0,0 +1,435 @@
...
+        TickmarksZLabels,
+    } );
+}
+
+template< QskLevelingSensorSkinlet::NodeRole >
+Q_REQUIRED_RESULT QRectF QskLevelingSensorSkinlet::subControlRect(
+    const QskLevelingSensor* const sensor, const QRectF& contentsRect ) const = delete;
+
+template<>
+Q_REQUIRED_RESULT QRectF QskLevelingSensorSkinlet::subControlRect< R::OuterDisk >(
```
peter-ha, line 145: I wonder if this template specialization is needed here

peter-ha (commented):
Just voicing my thoughts; Uwe might have a different idea on how this should be structured, or whether this should be a Qsk* class now etc.

src/controls/LevelingSensor/QskLevelingSensorSkinlet.cpp
rick-vogel: not needed, but subjectively nice to look at ;)

src/controls/LevelingSensor/QskLevelingSensorNodes.h
rick-vogel: Curiously Recurring Template Pattern

https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern

src/controls/LevelingSensor/QskLevelingSensorNodes.h
rick-vogel: - [ ] use clang-format on all changed files

playground/levelingsensor/Dashboard.cpp
rick-vogel: used and removed

src/controls/LevelingSensor/QskLevelingSensor.h
rick-vogel: Well AFAIK the copy-of-write semantics can coexist with this approach. Personally I think this siganture 100% communicates what is hapening: The caller gets a readonly reference. Any further processing and manipulation can then trigger copy-on-write semantics.

In my opinion this whole copy-on-write semantics makes it even more complicated. Especially when mixing QSK and Qt containers. As a caller I always have to ask myself if an container/object is being copied and when. While the const ref approach clearly defines it.

src/controls/LevelingSensor/QskLevelingSensor.cpp
rick-vogel: No, since the double case is not being used in this file. Still added it so future changes don't run into subtle narrowing type conversions.

src/controls/private/QskSGNodeUtility.h
```diff
@@ -0,0 +1,38 @@
...
+#include <QSGNode>
+
+namespace QskSGNode
+{
+    enum AppendMode
+    {
+        Sequential,
+        Recursive
+    };
+
```
rick-vogel, line 12: @uwerat This function is a candidate for public API and could be moved to `QskSGNode.h` next to `ensureNode`

src/controls/private/QskLevelingSensor.cpp
```diff
@@ -0,0 +1,178 @@
...
+  public:
+    QVector3D m_angle = { 45, 45, 45 };
+    Tickmarks m_tickmarks[ 3 ];
+    TickmarksLabels m_tickmarksLabels[ 3 ];
+    std::unordered_map< QskAspect::Subcontrol, QVector3D > m_subControlRotation;
+};
+
+QskLevelingSensor::QskLevelingSensor( QQuickItem* const parent )
+    : Inherited( parent )
+    , m_data( std::make_unique< QskLevelingSensor::PrivateData >() )
```
uwerat, line 76: What is the benefit of using std::make_unique here ?

If there is one we should change all the other classes, where the PrivateData pointer is initialized by a simple new.

src/controls/LevelingSensor/QskLevelingSensor.h
peter-ha: All that sounds good, but in all of Qt and QSkinny we return such classes by value, so IMO we should be consistent here too

uwerat: Well, for a user, who has been "socialized" from Qt APIs, there is a hidden message behind returning as reference: "the class being returned does not support copy-on-write".

Qt created their own type of containers back in the days, where there was no common STL and you could not rely on finding the same containers with each C++ compiler ( note that was also no std::string ). So it was actually mandatory for the cross platform concept. With Qt4 all containers have been redesigned from scratch - for some reasons ( I do not know ) not being replaced by the STL counter parts ( maybe because std::move did not yet exist ).

Unfortunately this lead to a situation, where a Qt addon project has to decide between the competing C++ and Qt concepts. For QSkinny I made the decision to go with Qt for the public API. The main reason for this decision is, that the majority of the QSkinny classes are derived from Qt classes and all parts of a class API should be consistent.

Considering designing Qsk classes, there is a simple rule of thumb: if a deep copy of a class is not cheap it is supposed to implement copy-on-write. Actually the idea of copy-on-write makes the life of a user simpler as he can copy around instances without thinking about extra costs.

Actually this worked perfectly fine until the range-based loops came up with C++11, where not so obvious "write" access of a container happens by accident.

src/controls/LevelingSensor/QskLevelingSensorNodes.h
uwerat: Please note that there are some pending changes around the axes/scales in this branch: https://github.com/uwerat/qskinny/tree/features/plots I did not merge them to master, because I'm not 100% happy with everything ( f.e aligning the lines to avoid the unwanted OpenGL antialiasing of vertical/horizontal lines ).

There is another branch https://github.com/uwerat/qskinny/tree/features/geometrynode, where I'm trying to introduce a QskGeometryNode to simplify the implementation of geometry nodes in combination with QSGFlatColorMaterial/QSGVertexColorMaterial.

I was distracted by the ongoing changes because of the Fluent2 skin - however it might be worth to check if there is something of value for your work in these branches.

src/controls/LevelingSensor/QskLevelingSensor.h
peter-ha: Maybe this and other information like the make_unique discussion is something we could gather on a list and publish it somewhere as coding guidelines or so?

src/controls/QskLevelingSensor.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
```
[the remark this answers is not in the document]
rick-vogel: done

src/controls/LevelingSensor/QskLevelingSensor.h
rick-vogel: done

src/controls/LevelingSensor/QskLevelingSensor.h
rick-vogel: Decision against virtual: setter and getter should not have more complexity and business logic that storing and returning values. Helper functions can be provided elsewhere.

peter-ha:
Not trying to bring pressure or so, but @uwerat do you know whether this will be a Qsk* class, or rather a class in an example or so? We are wondering how we should use/name this internally...

PR #339, opened 2023-08-08
QSK Radial Tickmarks Node

rick-vogel:
# Requirements

- Major, medium and minor tickmarks must be styleable
    - either by size in pixel or in %
- Tickmarks can be layouted as a circle or ellipse
- Tickmarks can be defined selectively
- Tickmarks can be aligned vertically centered, bottom or top
- Tickmarks line width can be styled
    - might not work for every backend, 1px is always supported

Recording 2023-08-08 at 15 05 15

### DIFF
--- a/playground/nodes/RadialNodes.cpp
+++ b/playground/nodes/RadialNodes.cpp
@@ -9,4 +9,15 @@ RadialTickmarks::RadialTickmarks( QQuickItem* const parent ) : QskControl( paren
     setStrutSizeHint( Lines, 2, 10 );
 }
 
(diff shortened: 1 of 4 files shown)


peter-ha:
First impression: Cool to have such a class, we are using it internally already.
- I would put the node into an own file, so it is easy to use it
- I guess we need QskArcMetrics to determine the angle(s)
- As an API I would have tried to copy the one from QskTickmarksNode, i.e. the update method (without orientation but with arc metrics etc.)
- I would call the examples different than nodes

Maybe Uwe can comment whether this class should be in QSkinny directly; then we could use it from at least the IOT dashboard (the light control to be precise).

PR #344, opened 2023-10-16
subwindows: Allow fading out

### LINKED ISSUE #10: Fading out of QskSubWindow
We need to find a way how to fade out QskSubWindow.

Classes derived from QskSubWindow are often dialogs and candidates for being created on the stack being processed in a recursive event loop. To solve this type of use case we might need to have some sort of "post mortem" handler grabbing the contents into an image or simply hijacking the scenegraph node subtree.

An obvious solution would be to let some sort of QskPopupManager let do the job.

peter-ha:
Resolves #10

### DIFF
--- a/skins/fluent2/QskFluent2Skin.cpp
+++ b/skins/fluent2/QskFluent2Skin.cpp
@@ -1877,6 +1877,8 @@ void Editor::setupSubWindow( const QskFluent2Theme& theme )
     using Q = QskSubWindow;
     const auto& pal = theme.palette;
 
+    setAnimation( Q::Panel | QskAspect::Position, 150 );
+
     setPadding( Q::Panel, { 0, 31, 0, 0 } );
     setBoxShape( Q::Panel, 7 );
     setBoxBorderMetrics( Q::Panel, 1 );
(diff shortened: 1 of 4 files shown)


peter-ha:
questions / thoughts:

- Right now the fading aspect is set to "Panel | QskAspect::Position", which doesn't make sense semantically, because we want to change the opacity and not the position. Shall we use a different Primitive or just leave out the Position?
- Do we always want to set the opacity? In the future there might also be the need to set a scale factor (see https://www.figma.com/file/NAWMapFlXnoOb86Q2H5GKr/Windows-UI-(Community)?type=design&node-id=2265-5171&mode=design&t=BBaBTz1yqM1M9p3A-0 and https://learn.microsoft.com/en-us/windows/apps/design/signature-experiences/motion), but this seems to regard windows and icons rather than dialogs. So I guess hardcoding the opacity is ok for now...
- For exact styling, we would also need to set easing curves as skin hints. I guess we don't want a setEasingCurve() in QskSkinHintTableEditor and rather use the generic setH
int() API, right? Again for now it seems we don't need this. 
- In order to style dialogs differently (e.g. critical might have another duration for fading in or out than warning, see link above), would we need those differentiations in QskDialog as variation or so? However the windows API (https://learn.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs) doesn't mention anything h
ere regarding different durations etc. for dialog types, so the whole motion API is not for dialogs apparently.

uwerat:
Better support for fading ( sliding ) in/out of popups has been implemented in QskPopup. QskMenu/QskSubWindow/QskDrawer are makeing use of it.

PR #345, opened 2023-10-20
Fluent2: Don't flatten border colors to a color without alpha value

peter-ha:
... or something along those lines.

before:

Screenshot from 2023-10-20 14-20-09

after:

Screenshot from 2023-10-20 14-19-45


uwerat:
Fix is "commited". However I didn't use the patch as it removes code, that is intended to test a different strategy how to resolve the grays. See the comment in the code I added with the commit.

PR #347, opened 2023-11-05
Add QskTickmarksMetrics / QskGraduationMetrics

vrcomputing:
Motivation: Just recently again I needed to be able to style the linear tickmarks sizes for all types (minor, medium, major), therefore  here a suggestion of how to make them configurable with the current proportional behavior.

**Advantages**

- Each tickmark type can freely be styled
- Tickmark type's length no longer implementation detail

image

### DIFF
--- a/src/common/QskFunctions.h
+++ b/src/common/QskFunctions.h
@@ -69,6 +69,11 @@ inline constexpr bool qskFuzzyCompare( qreal value1, qreal value2 )
     return qFuzzyCompare( value1, value2 );
 }
 
+inline constexpr bool qskConstrainedRatio( qreal ratio )
+{
+    return qBound( 0.0, ratio, 1.0 );
+}
+
 QSK_EXPORT qreal qskFuzzyFloor( qreal value, qreal stepSize );
 QSK_EXPORT qreal qskFuzzyCeil( qreal value, qreal stepSize );
 
--- a/src/common/QskFunctions.h
+++ b/src/common/QskFunctions.h
@@ -58,7 +58,7 @@ inline QMarginsF qskMargins( const QRectF& rect, const QRectF& innerRect )
         rect.bottom() - innerRect.bottom() );
 }
 
-inline bool qskFuzzyCompare( qreal value1, qreal value2 )
+inline constexpr bool qskFuzzyCompare( qreal value1, qreal value2 )
 {
     if ( qFuzzyIsNull( value1 ) )
         return qFuzzyIsNull( value2 );
--- a/src/common/QskFunctions.h
+++ b/src/common/QskFunctions.h
@@ -69,7 +69,7 @@ inline constexpr bool qskFuzzyCompare( qreal value1, qreal value2 )
     return qFuzzyCompare( value1, value2 );
 }
 
-inline constexpr bool qskConstrainedRatio( qreal ratio )
+inline constexpr qreal qskConstrainedRatio( qreal ratio )
 {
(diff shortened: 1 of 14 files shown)


src/common/QskFunctions.h
```diff
@@ -58,7 +58,7 @@ inline QMarginsF qskMargins( const QRectF& rect, const QRectF& innerRect )
         rect.bottom() - innerRect.bottom() );
 }
 
-inline bool qskFuzzyCompare( qreal value1, qreal value2 )
+inline constexpr bool qskFuzzyCompare( qreal value1, qreal value2 )
```
vrcomputing, line 61: @uwerat There are propably more functions here that can be `constexpr` as well

src/nodes/QskTickmarksMetrics.cpp
vrcomputing, line 1: Obsolete is header only implementation

src/nodes/QskTickmarksMetrics.h
```diff
@@ -0,0 +1,131 @@
...
+    constexpr void setMinorRatio( qreal ratio ) noexcept;
+
+    constexpr Q_REQUIRED_RESULT qreal operator[]( int index ) const noexcept;
+    constexpr Q_REQUIRED_RESULT qreal& operator[]( int index ) noexcept;
+
+  private:
+    qreal m_minorRatio = 0.7;
+    qreal m_mediumRatio = 0.85;
+    qreal m_majorRatio = 1.0;
+};
```
vrcomputing, line 47: @uwerat I was thinking about adding `Qt::SizeMode` but this would need more adaptions in the `QskTickmarksNode`. What do you think?

uwerat:
Guess QskTickmarkMetrics should be usable as skin hint:

a) some boilerplate code to make it usable with QVariant ( see qskRegisterIntervalF as an example )
b) interpolation needs to be implemented and registered in QskVariantAnimator.cpp

src/nodes/QskTickmarksMetrics.h
uwerat: Guess setting the length of the ticks as absolute values ( pixels/dp ) is what a user usually want to do.

IIRC the current implementation relates the tick lengths to a metric, that is given as a skin hint. IMHO this does not make much sense anymore as QskTickmarksMetrics allows to pass the tick lengths as skin hint instead.

uwerat:
Please note that I have been working on the tickmarks in the features/plots branch some time ago. I'm not 100% happy yet with what you find there, but it is much closer to the final result than what you find in master. ( f.e there is a QskScaleRenderer - no QskTickmarksNode anymore ).

Please split this merge request into a part that implements QskTickmarksMetrics and another that is using it. Then I can adjust features/plots also using it.

uwerat:
I would like to extend the concept of QskTickmarksMetrics to something, where you can configure more metrics of a scale. 
This might include the width of the ticks, if the scale has a backbone ( with/without arrow ) and maybe even options like the rotation/alignment of the tick labels. 

Having the lengths of the ticks is good enough for the first step, but I would like to start with a different terminology so that additional values can be added later without breaking the meaning of the class name.

Unfortunately the term "scale" is heavily used in the context of  transformations and "axis" might be a bit too specific for coordinate systems. But there seems to be another good one: "graduation" ( https://en.wikipedia.org/wiki/Graduation_(scale) )

So please rename QskTickmarksMetrics to QskGraduationMetrics.

uwerat:
QskGraduationMetrics is in master now

PR #354, opened 2023-12-21
Fix ninja multi config generator

vrcomputing:


.github/workflows/cmake.yml
```diff
@@ -104,7 +104,7 @@ jobs:
               cc: "gcc",
               cxx: "g++",
               archiver: "7z a",
-              generators: "Ninja",
+              generators: "Ninja Multi-Config",
```
vrcomputing, line 107: @uwerat this is meant as a temporary change but can be kept if we e.g. want to build and delivery multi config packages which I'm a fan of and qskinny is small enough that it doesn't matter size wise

PR #351, opened 2023-12-21
[WIP] Feature/CMake Install Paths

lgtmak:
* Updates the CMake `install()` calls to properly allow the installation of the libraries and header files as using best practices.
* Changes documentation to reflect modern install method

PR #355, opened 2023-12-22
Bug/Search for Qt6 then Qt5 before building

lgtmak:
There is a bug with find_package in which the ordering is not respected on Linux:

[Linux Bug](https://gitlab.kitware.com/cmake/cmake/-/issues/23575)

This makes the `qsk_setup_Qt` macro always load Qt5 (as opposed to suggested [practices](https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions)).

This will check for the Qt6 first then fallback to Qt5 if not found, thus replicating the best practice behavior.

To target a specific version pass `QSK_QT_VERSION` to the cmake build command: 

```sh
$ cmake -DQSK_QT_VERSION=Qt5 ../ && make
```

Documentation to be added in #351

### DIFF
--- a/cmake/QskFindMacros.cmake
+++ b/cmake/QskFindMacros.cmake
@@ -5,12 +5,11 @@
 
 macro(qsk_setup_Qt)
 
-    # Often users have several Qt installations on their system and
-    # need to be able to explicitly the one to be used. Let's see if
-    # standard cmake features are good enough or if we need to introduce
-    # something sort of additional option. TODO ...
-
-    find_package(QT "5.15" NAMES Qt6 Qt5 REQUIRED COMPONENTS Quick)
+    # Define a package QT, attempt Qt6, if not fallback to Qt5
+    find_package(QT NAMES Qt6 REQUIRED COMPONENTS Quick)
+    if (NOT QT_FOUND)
+        find_package(QT "5.15" NAMES Qt5 REQUIRED COMPONENTS Quick)
+    endif()
 
     if ( QT_FOUND )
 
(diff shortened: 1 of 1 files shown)


lgtmak:
Perhaps this is not the best approach and we should instead build the library for both Qt5 and Qt6?

lgtmak:
I am also now receiving this error while building:

```sh
-- Generating done (0.1s)
-- Build files have been written to: /home/hawk/Code/qskinny/build
[  0%] Built target qskinny_autogen_timestamp_deps
[  0%] Built target qskinny_autogen
[ 53%] Built target qskinny
[ 53%] Built target squiekskin_autogen_timestamp_deps
[ 54%] Built target squiekskin_autogen
[ 54%] Building CXX object skins/squiek/CMakeFiles/squiekskin.dir/QskSquiekSkinFactory.cpp.o
In file included from /home/hawk/Code/qskinny/skins/squiek/QskSquiekSkinFactory.cpp:33:
/home/hawk/Code/qskinny/build/skins/squiek/squiekskin_autogen/include/moc_QskSquiekSkinFactory.cpp:17:2: error: #error "This file was generated using the moc from 5.15.11. It"
   17 | #error "This file was generated using the moc from 5.15.11. It"
```

Not sure how this is possible.

lgtmak:
There is also two CMAKE vars that could be used instead:

‘CMAKE_FIND_PACKAGE_SORT_ORDER=NAME’

‘CMAKE_FIND_PACKAGE_SORT_DIRECTION=DEC’

lgtmak:
I decided on a hybrid approach, ready for review and merge.

PR #363, opened 2024-01-11
Unify QSkinny Cmake Package

vrcomputing:
- [x] fix qml export installation location
- [x] fix build break for non msvc compilers regarding pdbs


vrcomputing:
Renaming `QSkinnyTargets.cmake` to prevent confusion between framework and the `QSkinny` target deferred

lgtmak:
Please update the docs and examples for the namespace changes.

PR #370, opened 2024-01-19
Bug/CMAKE_PREFIX_PATH for find_package Qt Version

lgtmak:
This is a follow up to https://github.com/uwerat/qskinny/pull/355

and a response to: https://github.com/uwerat/qskinny/pull/364#issuecomment-1890442873

The build process will now use CMAKE_PREFIX_PATH to search for Qt versions but it still maintains the fix for proper sorting on Linux platforms to always build QSkinny at the intended targeted version.

### DIFF
--- a/cmake/QskFindMacros.cmake
+++ b/cmake/QskFindMacros.cmake
@@ -5,15 +5,12 @@
 
 macro(qsk_setup_Qt)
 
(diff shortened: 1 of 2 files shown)


lgtmak:
Nvm, did not realize this was resolved.

PR #373, opened 2024-01-19
Fix qskinnytools cmake

rick-vogel:
The `Qt` variable no longer exists

```
1>CMake Error at C:/repositories/qskinny_rveh-install-win/lib/cmake/QSkinny/QSkinnyTools.cmake:11 (add_custom_command):
1>  Error evaluating generator expression:
1>
1>    $<TARGET_FILE_DIR:::Svg>
1>
1>  No target "::Svg"
1>Call Stack (most recent call first):
1>  CMakeLists.txt:50 (qsk_svg2qvg)
```


cmake/QSkinnyTools.cmake
```diff
@@ -8,11 +8,18 @@
 function(qsk_svg2qvg SVG_FILENAME QVG_FILENAME)
     get_filename_component(QVG_FILENAME ${QVG_FILENAME} ABSOLUTE)
     get_filename_component(SVG_FILENAME ${SVG_FILENAME} ABSOLUTE)
+
+    if(TARGET Qt6::Svg)
+        set(QtSvgTarget Qt6::Svg)
+    elseif(TARGET Qt5::Svg)
+        set(QtSvgTarget Qt5::Svg)
+    endif()
```
rick-vogel, line 16: For some reason `$<TARGET_FILE_DIR:Qt::Svg>` doesn't work for `ALIAS`? Qt::Svg, but even if it would still break if the user configures `QT_NO_CREATE_VERSIONLESS_TARGETS` ... 😢

cmake/QSkinnyTools.cmake
```diff
@@ -8,11 +8,18 @@
...
+
     add_custom_command(
-        COMMAND svg2qvg ${SVG_FILENAME} ${QVG_FILENAME}
+        COMMAND Qsk::Svg2Qvg ${SVG_FILENAME} ${QVG_FILENAME}
```
rick-vogel, line 19: Hereby the `qsk_svg2qvg` is only usable from a consumer's perspective. Would need some mechanism to reuse it internally. Maybe even `ALIAS` would be the solution.

cmake/QSkinnyTools.cmake
```diff
@@ -8,11 +8,18 @@
...
         OUTPUT ${QVG_FILENAME}
         DEPENDS ${SVG_FILENAME}
-        WORKING_DIRECTORY $<TARGET_FILE_DIR:${Qt}::Svg>
+        WORKING_DIRECTORY $<TARGET_FILE_DIR:${QtSvgTarget}>
```
rick-vogel, line 22: Well this is also only half the truth. If the user built QSkinny with `BUILD_SVG2QVG_STANDALONE=OFF` the svg2qgv executable needs qskinny.so...

rick-vogel: @uwerat didn't we agree on removing `BUILD_SVG2QVG_STANDALONE` and making in always standalone?

uwerat: > @uwerat didn't we agree on removing `BUILD_SVG2QVG_STANDALONE` and making in always standalone?

Confirmed.

tools/svg2qvg/CMakeLists.txt
```diff
@@ -46,3 +46,28 @@ target_link_libraries(${target} PRIVATE Qt::Svg)
 set_target_properties(${target} PROPERTIES FOLDER tools)
 
 install(TARGETS ${target})
+
+# packaging
+set(PACKAGE_TARGET_FILENAME QSkinnyToolsTargets.cmake)
+
+set_target_properties(${target} PROPERTIES EXPORT_NAME Svg2Qvg)
+set_target_properties(${target} PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE)
```
rick-vogel, line 54: @uwerat I needed this in order to get svg2qvg running in my integration test on Mac. Does this cause problems from a user's perspective?

cmake/QSkinnyTools.cmake
```diff
@@ -8,11 +8,41 @@
...
+    # select platform specific wrapper script
+    if (CMAKE_SYSTEM_NAME MATCHES "Windows")
+        set(script ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/scripts/QSkinnySvg2Qvg.win.bat)
+    elseif (CMAKE_SYSTEM_NAME MATCHES "Darwin")
+        set(script ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/scripts/QSkinnySvg2Qvg.mac.sh)
+    elseif (CMAKE_SYSTEM_NAME MATCHES "Linux")
+        set(script ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/scripts/QSkinnySvg2Qvg.lin.sh)
+    else()
+        message(FATAL "Unsupported operating system")
+    endif()
```
rick-vogel, line 39: @uwerat Decided to do the platform specific stuff in script. Thereby not having to deal with custom environments, syntaxes, CMake % ; $ string escaping, CMake Generator specifics. Not beautifull but working for now...

PR #378, opened 2024-01-26
Build for WebAssembly including building everything statically etc.

### LINKED ISSUE #372: Webassembly
Is it possible to run QSkinny applications on the Webassemby platform: https://doc.qt.io/qt-6/wasm.html

peter-ha:
Resolves #372

### DIFF
--- a/cmake/QskBuildFunctions.cmake
+++ b/cmake/QskBuildFunctions.cmake
@@ -47,10 +47,16 @@ function(qsk_add_plugin target TYPE CLASS_NAME)
     # So for the moment better don't do:
     #      qt6_add_plugin(${target} SHARED ${CLASS_NAME} )
 
-    if(QT_VERSION_MAJOR VERSION_GREATER_EQUAL 6)
-        qt6_add_library(${target} SHARED )
+    if(BUILD_QSKDLL)
+        set(library_type SHARED)
     else()
-        add_library(${target} SHARED )
+        set(library_type STATIC)
+    endif()
+
+    if(QT_VERSION_MAJOR VERSION_GREATER_EQUAL 6)
+        qt6_add_library(${target} ${library_type})
+    else()
+        add_library(${target} ${library_type})
     endif()
 
     set_target_properties(${target} PROPERTIES
@@ -117,5 +123,11 @@ function(qsk_add_shaders target)
 
     qt6_add_shaders( ${target} "qskshaders" BATCHABLE PRECOMPILE QUIET
         PREFIX "/qskinny/shaders" ${ARGV} OUTPUTS ${outfiles} )
+
+    # pass on OUTPUT_TARGETS to the caller of this function
+    cmake_parse_arguments(PARSE_ARGV 1 arg "" "OUTPUT_TARGETS" "")
+    if (arg_OUTPUT_TARGETS)
+        set(${arg_OUTPUT_TARGETS} ${${arg_OUTPUT_TARGETS}} PARENT_SCOPE)
+    endif()
 
 endfunction()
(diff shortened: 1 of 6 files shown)


peter-ha:
- if BUILD_QSKDLL is not set, we build everything statically, including plugins. Is that ok or do we need another flag? 
- How about setting a global property LIBRARY_TYPE that is set to STATIC or SHARED depending on BUILD_QSKDLL? Then we could get rid of all the if(BUILD_QSKDLL) calls in the cmake files.

src/CMakeLists.txt
```diff
@@ -542,8 +539,14 @@ set_target_properties(${target} PROPERTIES
 # packaging
 set(PACKAGE_NAME ${PROJECT_NAME})
 
+set(QSK_EXPORTED_TARGETS ${target})
```
rick-vogel, line 542: I know I suggested it, but lets stick to local variables `lower_snake_case` and preserve `QSK_UPPER_CASE` for "global" variables. So please rename to e.g. `exported_targets`

src/nodes/QskTreeNode.cpp
```diff
@@ -5,7 +5,7 @@
 
 #include "QskTreeNode.h"
 
-static constexpr auto extraFlag =
+static auto extraFlag =
```
rick-vogel, line 8: why no longer `constexpr` ?

cmake/QskBuildFunctions.cmake
```diff
@@ -118,4 +124,9 @@ function(qsk_add_shaders target)
     qt6_add_shaders( ${target} "qskshaders" BATCHABLE PRECOMPILE QUIET
         PREFIX "/qskinny/shaders" ${ARGV} OUTPUTS ${outfiles} )
 
+    cmake_parse_arguments(PARSE_ARGV 1 arg "" "OUTPUT_TARGETS" "")
```
rick-vogel, line 127: Please add one liner comment on what happens here and why

rick-vogel (changes requested):
Just a little doc and renaming

cmake/QskBuildFunctions.cmake
peter-ha: added a comment; I used this line from qt6_add_shaders because we want to pass on OUTPUT_TARGETS to the caller of qsk_add_shaders

src/nodes/QskTreeNode.cpp
peter-ha: with constexpr we get this error:
```
/home/peter/dev/qskinny/src/nodes/QskTreeNode.cpp:9:5: error: integer value 33554432 is outside the valid range of values [0, 33554431] for the enumeration type 'Flag' [-Wenum-constexpr-conversion]
    9 |     static_cast< QSGNode::Flag >( QSGNode::IsVisitableNode << 1 );
      |     ^
1 error generated.
```
Any suggestion on how to fix this differently would be greatly appreciated :)

rick-vogel: I think this is a gray area of constexpressions not allowing undefined behavior vs. enumerator being only a set of constant on the underlying type ( e.g. int by deafult ).

Which compiler are you using?

Compromise: `static const auto extraFlag` thereby exiting `constexpr` land

https://stackoverflow.com/questions/66664841/undefined-behavior-of-constexpr-static-cast-from-int-to-scoped-enum-with-non-fix

rick-vogel: PS: https://eel.is/c++draft/dcl.enum#8

rick-vogel: PPS: see suggested

https://godbolt.org/z/n5WKde3ME

peter-ha: The compiler is clang 18.0 which comes from the Emscripten SDK (needed by WebAssembly). I changed the declaration to "static const auto"

rick-vogel (approved):

PR #390, opened 2024-02-22
push button: Update after setting the emphasis

peter-ha:
Otherwise we won't re-read the possibly new skin hints


uwerat:
Also added resetImplicitSize - the emphasis might also have an impact on the sizeHints

PR #394, opened 2024-02-22
Playground QskDataControlMapper

ovenpasta:
This new class resembles the equivalent in Qt Widgets of `QDataWidgetMapper`

It is useful when working with forms with database records with multiple fields.

One thing I've noticed is that QskTextInput & co. don't have the `Q_PROPERTY` option `USER true` for the value or text property, I've set them in this pull request but maybe not necessary.

I added a small example in the playground that could be further expanded with other goodies. I'm using some other controls in production like ListView and ComboBox with QAbstractItemModel support. Not sure how I could propose them here since they need some customization to the skins.


uwerat:
Your class seems to work with QObject - not only for QskControl.

What do you think about being more expressive with the class name. Model might be better than Data. I'm also not sure about "mapper" - the term "binding" has been established with QML and - since Qt6 - we also have property bindings for C++.

The class is inspired from  QDataWidgetMapper and its implementation will have similarities with the code from qdatawidgetmapper.cpp. However the code in this pull request is obviously copied in parts. Please replace these parts by your own implementation - otherwise the complete project would be affected by the rules of the GPL.

I created https://github.com/uwerat/qskinny/issues/395 for the missing USER attribute. Pull requests are welcome.

ovenpasta:
I completely forgot I copied some parts, sorry for that, was a few years ago. Now I rewrote most things and improved the code in various ways.

I renamed the class to QskModelObjectBinder... 

What do you think?

uwerat:
I did some adjustments and moved the code to a feature branch: https://github.com/uwerat/qskinny/tree/features/modelobjectbinder. ( I missed your last commit - please do it once more for the modified code ).

I also created a pull request for the feature branch ( https://github.com/uwerat/qskinny/pull/397 ) that is intended to have further discussions and/or reviews. Please rebase your working environment.

PR #397, opened 2024-02-25
Features/modelobjectbinder

uwerat:
QskModelObjectBinder introduced


src/controls/QskModelObjectBinder.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+    void updateObjectProperty( QObject* object, const Binding& binding )
+    {
+        const auto index = model->index( currentRowIndex.row(), binding.column );
+        binding.property.write( object, index.data( Qt::EditRole ) );
+    }
+
+  public:
+    QMap< QObject*, Binding > bindings;
+    QPointer< QAbstractItemModel > model;
```
uwerat, line 69: @ovenpasta do we need to store the row as QPersistentModelIndex. When using an int the row would be valid without the model ?

src/controls/QskModelObjectBinder.cpp
```diff
@@ -0,0 +1,174 @@
...
+    bindModel( model );
+}
+
+void QskModelObjectBinder::bindObject(
+    QObject* object, int column, const QByteArray& propertyName )
+{
+    if( object == nullptr )
+        return;
+
+    // does not work with dynamic properties ...
```
uwerat, line 91: @ovenpasta do we need dynamic properties ?

ovenpasta: Yes, consider the following snippet:
```
auto addRow = new QskPushButton("Add row at first row",v);
    connect(addRow,&QskPushButton::clicked,[=]() {
        auto r = table->record();
        r.setValue( 0, 0 );
        r.setValue(1, "new row");
        qDebug()<< "current row before insert" << mapper->currentRow();
        table->insertRecord( 0 ,r);
        qDebug()<< "current row after insert" << mapper->currentRow();
    });
```

Output:
```
current row before insert 0
current row after insert 1
```

So technically the currentRow will adjust to the new row index after the insert.

src/controls/QskModelObjectBinder.cpp
ovenpasta: Yes, consider the following snippet:
```
auto addRow = new QskPushButton("Add row at first row",v);
    connect(addRow,&QskPushButton::clicked,[=]() {
        auto r = table->record();
        r.setValue( 0, 0 );
        r.setValue(1, "new row");
        qDebug()<< "current row before insert" << mapper->currentRow();
        table->insertRecord( 0 ,r);
        qDebug()<< "current row after insert" << mapper->currentRow();
    });
```

Output:
```
current row before insert 0
current row after insert 1
```

So technically the currentRow will adjust to the new row index after the insert.

src/controls/QskModelObjectBinder.cpp
ovenpasta: Personally I'm not using them, let's see

ovenpasta:
there is something else missing, what if the model is destroyed?

uwerat:
@ovenpasta what about: https://doc.qt.io/qt-6/qdatawidgetmapper.html#orientation-prop

PR #412, opened 2024-05-26
fix issue with building with both examples and playground disabled

ovenpasta:
When building qskinny with both `-DBUILD_EXAMPLES=OFF `and `-DBUILD_PLAYGROUND=OFF`, cmake cannot find the QtQuick library. Probably it gets pulled as part of QuickWidgets and WebEngine

### DIFF
--- a/cmake/QskFindMacros.cmake
+++ b/cmake/QskFindMacros.cmake
@@ -24,6 +24,8 @@ macro(qsk_setup_Qt)
     else()
         message(FATAL_ERROR "Couldn't find any Qt package !")
     endif()
+    
+    find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Quick)
 
     # Using find_package is probably no good idea as all other Qt modules should
     # be located below ${_qt_cmake_dir}. Otherwise we might end up with modules


cmake/QskFindMacros.cmake
```diff
@@ -24,6 +24,8 @@ macro(qsk_setup_Qt)
     else()
         message(FATAL_ERROR "Couldn't find any Qt package !")
     endif()
+    
+    find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Quick)
```
uwerat, line 28: There is already a find_package( ... Quick) at the beginning of qsk_setup_Qt.
Could you please find out why it is not working in your case ?

ovenpasta: Yes it is strange, and the patch effectively fixes the issue in my case. I can reproduce both on windows / vcpkg / qt 6 and with Arch Linux / Qt 6.7.1.

I've found this:
[https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions](https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions)

They suggest that for older qt 5 versions should call find_package twice, one with the QT + NAMES and the second with the effective version. In my case it is needed even with Qt6

PR #413, opened 2024-05-29
Update Material3 Sliders

peter-ha:
Screenshot from 2024-05-29 16-57-06

Screenshot from 2024-05-29 16-57-09

### DIFF
--- a/src/controls/QskSlider.cpp
+++ b/src/controls/QskSlider.cpp
@@ -121,7 +121,15 @@ QSizeF QskSlider::handleSize() const
 
 QRectF QskSlider::handleRect() const
 {
-    return subControlRect( QskSlider::Handle );
+    auto rect = subControlRect( Handle );
+
+#if 1 // minimum handle strut size hardcoded here for now
+    const QSizeF strutSize( 60, 60 );
+    const auto w = qMax( ( strutSize.width() - rect.width() ) / 2, 0.0 );
+    const auto h = qMax( ( strutSize.height() - rect.height() ) / 2, 0.0 );
+#endif
+
+    return rect.marginsAdded( { w, h, w, h } );
 }
 
 void QskSlider::mousePressEvent( QMouseEvent* event )


peter-ha:
Some questions / comments:

* When handle is at beginning/end: Box shape doesn't work when rounding is bigger than bix width (or height). Any way this can be fixed in the box renderer?
* There is no ripple anymore in the M3 slider, but a value bubble instead. This seems to be very specific to M3; I wonder if we can do something about it like a M3SliderSkinlet, and put the design system specific things there? Haven't thought this through though...
* Re. stop indicators: This is implemented as series nodes, but since their color depends on whether they are in the fill or groove, we would need a new state, which feels wrong. For now I added 2 subcontrols (GrooveStopIndicators and FillStopIndicators), but we could also add a state, which would be just for styling those.
  Or should we also introduce something like QskSkinlet::sampleVariation(), similar to sampleStates(), which would just return NoVariation by default?
* I added a padding for the handle so it is easier to use with the finger (see the topmost commit). Otherwise the handle is very easy to miss for M3

uwerat:
The CI fix has been committed: https://github.com/uwerat/qskinny/commit/278fca26d7b4fd2dc810d610c6a2f6dccb68b1ec. Will have look at the slider code soon.

peter-ha:
> The CI fix has been committed: [278fca2](https://github.com/uwerat/qskinny/commit/278fca26d7b4fd2dc810d610c6a2f6dccb68b1ec). Will have look at the slider code soon.

Actually the fix was incomplete: It still fails for Qt 5 on Mac. The proper commit is (hopefully) this one: https://github.com/uwerat/qskinny/pull/413/commits/b32bd1ee0b5120839cd26837cb8832c5d9e6d7c3 , which also contains an explanation of the changes.

uwerat:
>     * When handle is at beginning/end: Box shape doesn't work when rounding is bigger than bix width (or height). Any way this can be fixed in the box renderer?

I'm afraid didn't get the problem completely. What metrics are rounded beyond which size and what is happening then ?

peter-ha:
note to self:

- Create issue for the box shape problem
- hardcode minimum strut size for the handle (for now)

peter-ha:
> * Create issue for the box shape problem

https://github.com/uwerat/qskinny/issues/415 has been created for the box shape issue

peter-ha:
> * hardcode minimum strut size for the handle (for now)

the newest commit of this branch hardcodes a minimum size for the handle

uwerat:
The various M3 slider configurations seem to be variations of features that IMO make sense for sliders in general.
I will try to generalize these features and to find implementations for fusion/fluent2.

I created a feature branch ( https://github.com/uwerat/qskinny/tree/features/slider ) from this pull request as a starting point for this investigation. Please start from this branch if you want to add something to the sliders.

Please start from this branch if you like to add something to your implementation from this pull request.

PR #431, opened 2024-10-08
gradients: Don't crash gammaray

peter-ha:
gammaray etc. reads the properties and when displaying the value calls the corresponding READ methods. This will lead to crashes when calling "wrong" methods for gradients.

### DIFF
--- a/src/common/QskGradient.cpp
+++ b/src/common/QskGradient.cpp
@@ -615,8 +615,6 @@ void QskGradient::setLinearDirection( const QskLinearDirection& direction )
 
 QskLinearDirection QskGradient::linearDirection() const
 {
-    Q_ASSERT( m_type == Linear );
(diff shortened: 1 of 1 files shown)


uwerat:
I'm not 100% convinced by this solution as gammaray shows a wrong information then ( of course this is better than running into an assertion ). Only one of the directions is valid and gammary should display this one only.

Some random ideas - not sure about them:

- maybe QML needs the WRITE method only and we could go without declaring the getters as READ method.
- maybe we can define the property in a way, that gammaray skips it
- maybe there is a way to tell gammary to check the type() first.
- maybe it is possible to replace the 3 properties by 1 property "direction", that is a QVariant.
- ...

uwerat:
assertions have been weakened to warnings,

PR #440, opened 2024-10-20
spin box: Improve for M3

### LINKED ISSUE #398: QskSpinBox Buttons for Material3
Material3 does not have any spin boxes and we do not have a specification how to display it.

However other toolkits supporting Material3 display the buttons in a different way as we do at the moment:

- https://www.qt.io/blog/material-3-changes-in-qt-quick-controls
- https://pub.dev/packages/flutter_spinbox

As long as we do not know better we should align our implementation with what others are doing:

- The Up/Down buttons should use '-'/'+' symbols
- The Up/Down buttons should look like a QskPushButton with  QskPushButton::VeryLowEmphasis

peter-ha:
Resolves #398

normal:

Screenshot from 2024-10-20 17-43-59

hovered:

Screenshot from 2024-10-20 17-43-43

pressed:

Screenshot from 2024-10-20 17-44-03

Comments:
- size of up/down buttons: normal push button size (i.e. height of 40 dp)
- emphasis: As mentioned in the ticket emphasis is like the very low button emphasis
- spin box height has been chosen to be the same as the menu height (48 dp). Different
  controls have different heights here, e.g. outlined text field has 56 dp and chips
  only 32 dp. So we could also chose another height here.
- size of icons: Same as push button icons.
  (Panel size is 40 dp, icon padding is 11 dp, so icon size is 40-2*11=18 dp)

### DIFF
--- a/designsystems/material3/QskMaterial3Skin.cpp
+++ b/designsystems/material3/QskMaterial3Skin.cpp
@@ -898,6 +898,7 @@ void Editor::setupSpinBox()
 
     setHint( Q::Panel | QskAspect::Style, Q::ButtonsLeftAndRight );
 
+    setStrutSize( Q::Panel, -1.0, 48_dp );
     setBoxShape( Q::Panel, 4_dp );
     setBoxBorderMetrics( Q::Panel, 1_dp );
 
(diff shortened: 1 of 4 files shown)


uwerat:
I adjusted the code to make use of QCursor::pos() - instead of storing the position as skin hint

peter-ha:
Commits for reference: 932d34aea8eb7d4ca727fca5a57898ed22721277 and b7c2b2769127e299b6e177fd8bae4a213f7b8367

PR #460, opened 2024-11-04
progress indicators: Update to newest M3 spec

peter-ha:
Screenshot from 2024-11-04 13-58-45

I had to introduce a distinction for both progress bars and progress rings whether the groove is one piece or two:

- If it is one piece (F2 and Fusion), we want to draw the full groove below the fill, otherwise it would look strange: (The screenshot below shows the strange behavior that is being avoided with this PR).

Screenshot from 2024-10-29 09-16-05

- If it is two pieces, we need to have a spacing between groove and fill, see the first screenshot above, which we cannot achieve with only one groove bar or arc.


uwerat:
Functionality has been merged - however not 1:1 as being proposed by this pull reauest

PR #471, opened 2024-11-12
Fix crash with QskSlider in gallery on Windows

peter-ha:


src/controls/QskSlider.cpp
```diff
@@ -32,7 +32,7 @@ class QskSlider::PrivateData
     QPointF pressedPos;
     qreal pressedValue;
     bool tracking : 1;
-    Qt::Orientation orientation : 2;
+    Qt::Orientation orientation : 3;
```
uwerat, line 35: Why do we need 3 bits ?

src/controls/QskSliderSkinlet.cpp
```diff
@@ -136,15 +136,17 @@ QRectF QskSliderSkinlet::sampleRect(
...
+        r.setLeft( qMax( 0.0, x ) );
     }
     else
     {
-        r.setLeft( r.center().x() - size.width() / 2 );
+        const auto x = r.center().x() - size.width() / 2;
+        r.setLeft( qMax( 0.0, x ) );
         const auto y = r.bottom() - slider->valueAsRatio( pos ) * r.height() - size.height() / 2;
-        r.setTop( y );
+        r.setTop( qMax( 0.0, y ) );
```
uwerat, line 149: From the title of this pull request I guess it is about preventing a situation, where the application crashes. Wouldn't it be better to add defensive check where the crash happens instead of avoiding running into that code path ?

src/controls/QskSlider.cpp
peter-ha: First, Qt::Horizontal and Qt::Vertical are defined as 1 and 2, not 0 and 1.
Then enums in Visual Studio C++ apparently are signed (as opposed to with gcc and clang), which means that Qt::Vertical (10) is interpreted as -2 and not 2.

So the following code on Windows:
```
QSlider s(Qt::Vertical);
qDebug() << s.orientation();
```
... produces this output:
```
Qt::Orientation(-2)
```

In light of this I am not sure using bit fields for enumerations is worth it at all...

src/controls/QskSliderSkinlet.cpp
peter-ha: Yeah apparently the crash happens when y is < 0, which is calculated in the line above.
This is triggered by the orientation being neither Vertical nor Horizontal, hence the size hint is 0. In theory it can still happen that y is < 0 with layouting etc; do you know how a defensive check could look like?

src/controls/QskSlider.cpp
uwerat: Ah of course - it has to be "uint orientation : 2;"

src/controls/QskSliderSkinlet.cpp
uwerat: There should be no crash only because assigning a value < 0 to a local variable. Guess the crash is because of processing the calculated rectangle somewhere else. Then the defensive check would be against the questionable rectangle, where the crash happens.

Actually it might be possible, that a scene graph ends up being outside of the bounding rectangle of the item - f.e when not having a proper layout code. Another hypothetical situation I can imagine is when working with transformations.

src/controls/QskSlider.cpp
uwerat: I agree, that this optimization is not that important - when not having more than one bit field it is even of no value. However it only requires one cast in the getter and isn't hard to achieve.

I noticed, that we have the same problem in QskPageIndicator.

uwerat: I fixed the bitfield issues in QskPageIndicator and QskSlider.
The reason for the crash - and how to avoid it best - remains open.

src/controls/QskSliderSkinlet.cpp
uwerat: I recently fixed a heap corruption from the box renderer: https://github.com/uwerat/qskinny/commit/b5c56f7f8043391d1236eefcf4510d54d9aee4a2

Maybe this was also the reason for your crash ?

peter-ha: > I recently fixed a heap corruption from the box renderer: [b5c56f7](https://github.com/uwerat/qskinny/commit/b5c56f7f8043391d1236eefcf4510d54d9aee4a2)
> 
> Maybe this was also the reason for your crash ?

Yes, it turns out with this patch the crash does not appear anymore.

PR #502, opened 2025-01-10
Say hello to QskTextEdit

### LINKED ISSUE #109: QskTextEdit
A similar wrapper like QskTextInput needs to be implemented

peter-ha:
Resolves #109

### DIFF
--- a/src/controls/QskTextAreaSkinlet.cpp
+++ b/src/controls/QskTextAreaSkinlet.cpp
@@ -1,0 +1,103 @@
+/******************************************************************************
+ * QSkinny - Copyright (C) The authors
+ *           SPDX-License-Identifier: BSD-3-Clause
+ *****************************************************************************/
+
+#include "QskTextAreaSkinlet.h"
+#include "QskTextArea.h"
+
+using Q = QskTextArea;
+
+QskTextAreaSkinlet::QskTextAreaSkinlet( QskSkin* skin )
+    : Inherited( skin )
+{
+    setNodeRoles( { PanelRole, TextPanelRole, PlaceholderTextRole } );
+}
+
+QskTextAreaSkinlet::~QskTextAreaSkinlet()
+{
+}
+
+QRectF QskTextAreaSkinlet::subControlRect( const QskSkinnable* skinnable,
+    const QRectF& contentsRect, QskAspect::Subcontrol subControl ) const
+{
+    if ( subControl == Q::Panel )
+        return contentsRect;
+
+    if ( subControl == Q::TextPanel )
+        return skinnable->subControlContentsRect( contentsRect, Q::Panel );
+
+    if ( subControl == Q::PlaceholderText )
+    {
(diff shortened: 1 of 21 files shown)


peter-ha:
Screenshot from 2025-01-10 15-20-47

Screenshot from 2025-01-10 15-20-49

Screenshot from 2025-01-10 15-20-53

Screenshot from 2025-01-10 15-20-55

Screenshot from 2025-01-10 15-20-59

Screenshot from 2025-01-10 15-21-02

src/controls/QskTextAreaSkinlet.cpp
```diff
@@ -0,0 +1,103 @@
...
+            if ( text.isEmpty() )
+                return nullptr;
+
+            const auto subControl = Q::PlaceholderText;
+
+            QskSkinHintStatus status;
+
+            auto options = skinnable->textOptionsHint( subControl, &status );
+            if ( !status.isValid() )
+                options.setElideMode( Qt::ElideRight );
```
uwerat, line 66: Does the elide mode make much sense here ? A text area is multi line and I would expect, that the placeholder text would be fitted into the viewport. ( -> QskTextOptions::Wrap )

uwerat: But wait - you are probably right. One also wouldn't expect to have to scroll the placeholder text.

src/controls/QskTextEdit.h
```diff
@@ -0,0 +1,202 @@
...
+        ActivationOnKey = 1 << 2,
+
+        ActivationOnInput = ActivationOnMouse | ActivationOnKey,
+        ActivationOnAll = ActivationOnFocus | ActivationOnMouse | ActivationOnKey
+    };
+
+    Q_ENUM( ActivationMode )
+    Q_DECLARE_FLAGS( ActivationModes, ActivationMode )
+
+    enum EchoMode
```
uwerat, line 71: do we need this mode here ?

peter-ha: Right, that was copied over from QskTextInput, I just removed it.

uwerat:
Code is in master now - beside the design system specific skinlets ( do we have any specific decorations ? ).

uwerat:
Don't believe, that we need an error state - at least suggest not to implement it for the first release.

uwerat:
I also noticed, that the API of QQuickTextInput is not exported completely. But I thought I will have a second look at them once I'm done with QskTextField and friends. F.e I wouldn't add undo/redo to a password field and where to export the related API depends on the type of controls we will have ...

uwerat:
I noticed, that it is not possible to enter newlines with the keyboard ?

peter-ha:
> Code is in master now - beside the design system specific skinlets ( do we have any specific decorations ? ).

Isn't that similar to the ones from the text input? I.e. for M3 we would have the label text in the border or so, for which we would need an own skinlet? I couldn't even find a text edit for M3, but I assumed we would implement it in analogy to the text fields.

PR #507, opened 2025-01-30
implement dialogs for selection of file / directory / color / font

peter-ha:
## general:

* I introduced some private (non-exported) headers for the dialogs; I wonder whether we want to introduce a private folder to differentiate between exported and not exported classes?
* The templated class WindowOrSubwindow allows for the code to be written only once, see e.g. the class FileSelectionWindow. Otherwise we would have to duplicate the code like we do in QskMessage(Sub)Window and QskSelection(Sub)Window.
* For the above class I needed to copy some functions from QskDialog.cpp; maybe we should have them in a common place somewhere.
* I also copied code from Qsk(Sub)Window into QskWindowOrSubWindow; we should also decide how to deal with that.

## file / directory dialogs:

* In the file list view we needed to differentiate between single clicks and double clicks
* Two functions were copied from other parts and could be moved to QskFunctions or so:
    - qskRowAt() from QskListView.cpp
    - qskClonedMouseEvent() from QskGestureRecognizer.cpp

## color picker dialog:

* There was some work by Rick using shaders at https://github.com/vrcomputing/qskinny/tree/colorpicker .
  For now this has been implemented in software, i.e. filling the QImage pixel by pixel.

### DIFF
--- a/examples/gallery/dialog/DialogPage.cpp
+++ b/examples/gallery/dialog/DialogPage.cpp
@@ -107,6 +107,9 @@ namespace
             auto colorSelectionButton = new Button( "Color selection", this );
             connect( colorSelectionButton, &Button::clicked, this, &ButtonBox::execColorSelection );
 
+            auto fontSelectionButton = new Button( "Font selection", this );
+            connect( fontSelectionButton, &Button::clicked, this, &ButtonBox::execFontSelection );
+
             setExtraSpacingAt( Qt::BottomEdge );
         }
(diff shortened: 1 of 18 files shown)


peter-ha:
Screenshot from 2025-01-30 13-42-40
Screenshot from 2025-01-30 13-42-48
Screenshot from 2025-01-30 13-42-55
Screenshot from 2025-01-30 13-43-14

PR #532, opened 2026-03-19
text renderer: Use same methods for calculating text size and text re…

### LINKED ISSUE #519: text renderer: textSize() and actual rendering size differ
I hit a case with rendering multiple lines where querying QskTextRenderer::textSize() first and then rendering text into a rect with the size given by that function.
Sometimes this did not match, i.e. the rendered size had a bigger height than the textSize() due to an additional like break. Maybe this is because textSize() uses QFontMetricsF and the function used for rendering uses QTextLayout.
Maybe the solution would be to call qskLayoutText() from textSize() to be sure the returned sizes always match.

P.S.: Will try to provide a reproducible example soon™

peter-ha:
…ndering

Resolves #519

### DIFF
--- a/src/nodes/QskPlainTextRenderer.cpp
+++ b/src/nodes/QskPlainTextRenderer.cpp
@@ -18,23 +18,6 @@ QSK_QT_PRIVATE_END
 
 #define GlyphFlag static_cast< QSGNode::Flag >( 0x800 )
 
-QSizeF QskPlainTextRenderer::textSize(
-    const QString& text, const QFont& font, const QskTextOptions& options )
-{
-    // result differs from QQuickText::implicitSizeHint ???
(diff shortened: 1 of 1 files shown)


peter-ha:
Problem with this approach: QFontMetricsF might return different values. We have a case internally where QFontMetricsF::horizontalAdvance() return 60px, but the renderer is using 61px, so the last character is cut off.

QFontMetricsF is using QTextEngine directly, while QskTextRenderer uses the more high level QTextLayout.

We could try looking into how QTextLayout is using QTextEngine and try to mimic the behavior...

uwerat:
Branch has been used as starting point for https://github.com/uwerat/qskinny/tree/features/plaintextrenderer.
Needs more testing

